PR #113, opened 2015-02-13
rxnetty-0.4.6

brharrington:
Fixes bug that can lead to many spurious metrics if
unnamed clients are used.


brharrington:
Looks like maven central finally shows the new version. Re-running travis build.

PR #114, opened 2015-02-17
Change to new netflixoss build

rspieldenner:
If you guys want the new build stuff. We can talk about how to use travis with this (reactivex is).

If you decide to take the pull request add a v0.19.0 tag to where 0.19 was released.


brharrington:
Summarizing conversation with @rspieldenner about the implications: 

* Version will now be managed with tags on the git repo rather than the `gradle.properties` file.
* Dependencies in the pom will be marked as runtime instead of compile. This has caused some issues with some other projects (e.g. https://github.com/Netflix/ribbon/issues/206), we'll revisit if it becomes an issue here.
* We'll automatically get cloudbees jobs, but we can disable them and just stick with travis.

PR #120, opened 2015-02-20
Disable cloudbees pullrequest job creation/existence

rspieldenner:
This will cause our job DSL to not create and also remove the cloudbees pull request job.


brharrington:
Thanks!

PR #122, opened 2015-02-23
use http helper from iep-rxhttp

brharrington:
The http helper classes were moved to a separate
lib as there was some interest in using them without
the spectator nflx plugin.


brharrington:
iep libs were accidently built targeting java 8. A new version targeting java 7 is being built.

PR #169, opened 2015-06-24
Make the TagList.create methods publicly accessible.

bondj:
It would useful to be able to use the static create methods in TagList.


brharrington:
Talking in person, the use-case can be solved with one of the `Id.withTags` calls:

http://netflix.github.io/spectator/javadoc/HEAD/spectator-api/com/netflix/spectator/api/Id.html#withTags-java.util.Map-

PR #177, opened 2015-07-10
Change the behavior of rollup method to use first/newest value for du…

pstout:
…plicated tag names.

Add unit test to verify the behavior.

### DIFF
--- a/spectator-api/src/main/java/com/netflix/spectator/api/DefaultId.java
+++ b/spectator-api/src/main/java/com/netflix/spectator/api/DefaultId.java
@@ -64,17 +64,18 @@ public final class DefaultId implements Id {
   }
 
   /**
-   * Returns a new id with the tag list sorted by key and with no duplicate keys. If a duplicate
-   * is found the last entry in the list with a given key will be used.
+   * Returns a new id with the tag list sorted by key and with no duplicate keys.  This is equivalent to
+   * {@code rollup(Collections.<String>emptySet(), false)}.
    */
   public DefaultId normalize() {
     return rollup(EMPTY, false);
   }
 
   /**
-   * Create a new id by removing tags from the list. This operation will 1) sort the list by the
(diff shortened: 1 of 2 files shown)


spectator-api/src/main/java/com/netflix/spectator/api/DefaultId.java
```diff
@@ -64,17 +64,18 @@ public DefaultId(String name) {
   }
 
   /**
-   * Returns a new id with the tag list sorted by key and with no duplicate keys. If a duplicate
-   * is found the last entry in the list with a given key will be used.
+   * Returns a new id with the tag list sorted by key and with no duplicate keys.  This is equivalent to
+   * {@code rollup(Collections.<String>emptySet(), false)}.
    */
   public DefaultId normalize() {
     return rollup(EMPTY, false);
```
brharrington: Not sure why I didn't actually use `Collections.<String>emptySet()` here.

spectator-api/src/main/java/com/netflix/spectator/api/DefaultId.java
```diff
@@ -64,17 +64,18 @@ public DefaultId(String name) {
...
   }
 
   /**
-   * Create a new id by removing tags from the list. This operation will 1) sort the list by the
-   * tag keys, 2) dedup entries if multiple have the same key, and 3) remove keys specified by
-   * the parameters.
+   * Create a new id by possibly removing tags from the list.  This operation will:<br/>
+   *     1) remove keys as specified by the parameters<br/>
+   *     2) dedup entries that have the same key, the first value associated with the key will be the one kept,<br/>
+   *     3) sort the list by the tag keys.
```
brharrington: I think the recommended way of doing this in javadoc is to use the html `<ol>` tags. However, this seems to look fine in the javadoc output and I think this is easier to read in the code which I do more than looking at the html.

From a usage standpoint, do you think it will be clear to users that the the first value associated with the key will be the last value added using `withTag(k, v)`?

PR #182, opened 2015-08-04
use registry for creating the ids in t-digest lib

brharrington:
Updates the Json utility class to use an injected registry
to create the ids rather than using `DefaultId` directly.

This change does break backwards compatibility of some of
the classes to pass in the registry. However, it is still
experimental and our use of those classes is only via tests
or via a guice module so it will not break common usage.

cc @pstout

### DIFF
--- a/spectator-reg-tdigest/src/main/java/com/netflix/spectator/tdigest/StreamTDigestReader.java
+++ b/spectator-reg-tdigest/src/main/java/com/netflix/spectator/tdigest/StreamTDigestReader.java
@@ -15,6 +15,8 @@
  */
 package com.netflix.spectator.tdigest;
 
+import com.netflix.spectator.api.Registry;
+
 import java.io.DataInputStream;
 import java.io.IOException;
 import java.io.InputStream;
@@ -27,17 +29,19 @@ import java.util.List;
  */
 public class StreamTDigestReader implements TDigestReader {
 
+  private final Json json;
   private final DataInputStream in;
 
   private final byte[] buf = new byte[TDigestWriter.BUFFER_SIZE];
 
   /** Create a new instance. */
-  public StreamTDigestReader(InputStream in) {
-    this.in = new DataInputStream(in);
+  public StreamTDigestReader(Registry registry, InputStream in) {
+    this(registry, new DataInputStream(in));
   }
 
   /** Create a new instance. */
-  public StreamTDigestReader(DataInputStream in) {
+  public StreamTDigestReader(Registry registry, DataInputStream in) {
+    this.json = new Json(registry);
     this.in = in;
   }
 
@@ -53,7 +57,7 @@ public class StreamTDigestReader implements TDigestReader {
       if (length != size) {
         throw new IOException("unexpected end of stream");
       }
-      return Json.decode(buf, 0, length);
+      return json.decode(buf, 0, length);
     }
   }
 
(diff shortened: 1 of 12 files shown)


spectator-reg-tdigest/src/main/java/com/netflix/spectator/tdigest/StreamTDigestReader.java
```diff
@@ -65,9 +69,10 @@ public StreamTDigestReader(DataInputStream in) {
    * Consume all data from the stream create a list of the measurement batches as they were
    * read from the stream.
    */
-  static List<List<TDigestMeasurement>> readAll(InputStream in) throws IOException {
+  static List<List<TDigestMeasurement>> readAll(Registry registry, InputStream in)
```
pstout: Should this method be made non-static and use the registry passed into the class constructor?  The constructor would need to be changed to keep a direct reference to the Registry and not just use it to construct the Json object.

spectator-reg-tdigest/src/main/java/com/netflix/spectator/tdigest/StreamTDigestReader.java
```diff
@@ -27,17 +29,19 @@
...
 
+  private final Json json;
   private final DataInputStream in;
 
   private final byte[] buf = new byte[TDigestWriter.BUFFER_SIZE];
 
   /** Create a new instance. */
-  public StreamTDigestReader(InputStream in) {
-    this.in = new DataInputStream(in);
+  public StreamTDigestReader(Registry registry, InputStream in) {
```
pstout: Should this take a Json object rather than a Registry?

spectator-reg-tdigest/src/main/java/com/netflix/spectator/tdigest/TDigestWriter.java
```diff
@@ -43,6 +44,13 @@
   private final ByteBuffer buf = ByteBuffer.allocate(BUFFER_SIZE);
   private final ByteBufferOutputStream out = new ByteBufferOutputStream(buf, 2);
 
+  private final Json json;
+
+  /** Create a new instance. */
+  TDigestWriter(Registry registry) {
```
pstout: Should this take a Json object rather than a Registry?

spectator-reg-tdigest/src/main/java/com/netflix/spectator/tdigest/StreamTDigestReader.java
brharrington: I think it can change to just a `List<List<TDigestMeasurement>> readAll()` method that uses the stream. I would prefer to do that as a separate PR though.

spectator-reg-tdigest/src/main/java/com/netflix/spectator/tdigest/StreamTDigestReader.java
brharrington: This class is public, the `Json` object is package private. I didn't want to expose that class outside the package.

spectator-reg-tdigest/src/main/java/com/netflix/spectator/tdigest/StreamTDigestReader.java
brharrington: Addressed in: https://github.com/Netflix/spectator/pull/183

PR #210, opened 2015-08-28
Clarify TagList.prepend behavior

pstout:


pstout:
@brharrington This passes all of the API tests except for the CompatibilityTest which breaks due to some changes in the order of the tags in the produced strings.  I didn't want to update that test until you had a chance to comment.

brharrington:
It should be ok to go ahead and update the expected list. I should have normalized before emitting the output, but with this change that should no longer be needed.

brharrington:
Test fixed and commits squashed in #220

PR #212, opened 2015-09-03
Adds an extension that collects AWS SDK metrics into the Spectator Registry

cfieber:
@bpitman - this is what I was talking about earlier today


spectator-ext-aws/src/main/java/com/netflix/spectator/aws/SpectatorServiceMetricCollector.java
```diff
@@ -0,0 +1,101 @@
...
+    public SpectatorServiceMetricCollector(Registry registry) {
+        super();
+        this.registry = Preconditions.checkNotNull(registry, "registry");
+    }
+
+    @Override
+    public void collectByteThroughput(ByteThroughputProvider provider) {
+        if (provider.getByteCount() > 0) {
+            registry.counter(getThroughputBytesId(provider)).increment(provider.getByteCount());
+            registry.timer(getThroughputDurationId(provider)).record(provider.getDurationNano(), TimeUnit.NANOSECONDS);
```
brharrington: It seems like the duration is so you can work out the rate for a specific request. The counter will report bytes/second for the overall throughput and there is a separate latency stat. I'm not sure what this timer is adding. Is it useful for something?

spectator-ext-aws/src/main/java/com/netflix/spectator/aws/SpectatorServiceMetricCollector.java
```diff
@@ -0,0 +1,101 @@
...
+        final long durationNanos = toNanos(provider.getDurationMilli());
+        registry.timer(getServiceLatencyId(provider)).record(durationNanos, TimeUnit.NANOSECONDS);
+    }
+
+    private Id getServiceLatencyId(ServiceLatencyProvider provider) {
+        return registry.createId(getServiceLatencyMetric(), getServiceLatencyTags(provider));
+    }
+
+    private String getServiceLatencyMetric() {
+        return "AWS_Latency";
```
brharrington: The naming should follow the same conventions as other spectator libs. Though I noticed I don't have them documented on the wiki, so I'll take that as an action item. You can look at some of the other examples:

https://github.com/Netflix/spectator/wiki/Memory-Pools#jvmmemoryused

I think there is an internal doc with some details I can point you at also.

spectator-ext-aws/src/main/java/com/netflix/spectator/aws/SpectatorServiceMetricCollector.java
```diff
@@ -0,0 +1,101 @@
...
+        if (provider.getByteCount() > 0) {
+            registry.counter(getThroughputBytesId(provider)).increment(provider.getByteCount());
+            registry.timer(getThroughputDurationId(provider)).record(provider.getDurationNano(), TimeUnit.NANOSECONDS);
+        }
+
+    }
+
+    @Override
+    public void collectLatency(ServiceLatencyProvider provider) {
+        final long durationNanos = toNanos(provider.getDurationMilli());
```
brharrington: Maybe add a comment that `provider.getDurationMilli()` returns a double and needs to be converted to nanos to avoid losing precision. When I first looked at this I was wondering why it didn't use use `TimeUnit.MILLISECONDS`.

spectator-ext-aws/src/main/java/com/netflix/spectator/aws/SpectatorRequestMetricCollector.java
```diff
@@ -0,0 +1,81 @@
...
+        this.registry = Preconditions.checkNotNull(registry, "registry");
+    }
+
+    @Override
+    public void collectMetrics(Request<?> request, Response<?> response) {
+        final AWSRequestMetrics metrics = request.getAWSRequestMetrics();
+        if (metrics.isEnabled()) {
+            final String[] tags = getTags(request, response);
+            metrics.getTimingInfo().getAllCounters().forEach((name, count) -> {
+                registry.gauge(registry.createId(counterName(name), tags), count);
```
brharrington: Registering a gauge for each request seems odd. Gauges are used to provide a hook so the registry can query the value of something else that is typically going to be around for a while. For example the number of active threads in a thread pool.

Maybe this should be a counter, but I would have to look at how they are using it. The TimingInfo javadocs aren't much help.

brharrington:
Thanks. I'm going to pull it into a local branch so I can see the data it generates.

spectator-ext-aws/src/main/java/com/netflix/spectator/aws/SpectatorServiceMetricCollector.java
cfieber: I'm not sure - I haven't found the part of the SDK that reports this (guessing it is something more like S3 to track file transfers)

I'll kill off the timer

spectator-ext-aws/src/main/java/com/netflix/spectator/aws/SpectatorRequestMetricCollector.java
cfieber: The counter values do seem to be more long-lived stats around the underlying Apache HttpClient resources, and the TimingInfo counter API supports both a set and increment so definitely not monotonically increasing.

It might actually be a combination of values that would be gauge-like vs counter-like.

I was trying to avoid adding any knowledge or expectation about the names in here so it wouldn't have to care about or be updated when or if the SDK changed, but there is a MetricType enum (AWSRequestMetrics.Field) - I could potentially just pick the set of items worth tracking out of there and then explicitly pick those out in here instead of scraping the whole counter and timing metric collections

spectator-ext-aws/src/main/java/com/netflix/spectator/aws/SpectatorServiceMetricCollector.java
cfieber: I'm actually not even sure how valid this whole class is - the providers passed as collect arguments don't seem to expose enough info to add enough tags for these metrics.

For example: there is nothing indicating the region for the service, so collecting these is just going to be a mashup of the service latency to each region (maybe useful if you only talk inside a region, not so much if you have a service that makes calls to multiple regions)

The provider's all have a providerId attribute but I dropped that from the tags since it turned out to just be the default Object::toString value

spectator-ext-aws/src/main/java/com/netflix/spectator/aws/SpectatorRequestMetricCollector.java
brharrington: >  so it wouldn't have to care about or be updated when or if the SDK changed

We have been burned by doing that several times in the past. It can lead to values getting reported in that are quite confusing and everyone trying to consume the data now needs to go back and understand what it means. If a tradeoff needs to be made I think we should make it easy for someone consuming the data and hard for the person reporting it.

cfieber:
@brharrington rejigged this fairly significantly if you wouldn't mind taking another look. Also would you like it squashed?

spectator-ext-aws/src/main/java/com/netflix/spectator/aws/SpectatorRequestMetricCollector.java
```diff
@@ -0,0 +1,197 @@
...
+        public String getName() {
+            return name;
+        }
+
+        public Optional<String> getValue(AWSRequestMetrics metrics) {
+            return firstValue(metrics.getProperty(field), tagExtractor);
+        }
+    }
+
+    private static String decapitalize(String s) {
```
brharrington: Why not use java.beans.Introspector.decapitalize?

spectator-ext-aws/README.md
```diff
@@ -0,0 +1,133 @@
...
+---------------------------------|-----------
+aws.request.bytesProcessed       | AWSRequestMetrics.Field.BytesProcessed,
+aws.request.httpClientRetryCount | AWSRequestMetrics.Field.HttpClientRetryCount,
+aws.request.requestCount         | AWSRequestMetrics.Field.RequestCount
+
+The following request metrics are captured as timers:
+
+metric name                      | SDK metric
+---------------------------------|-----------
+# spectator-ext-aws
```
brharrington: Looks like this part was duplicated from the top.

brharrington:
I'm not too concerned with squashing the commits. The comments provide enough context for someone to follow the log.

I may still want to tweak how some of the data is reported, but it will be easier to see if that is needed by collecting some data. So I'm inclined to merge and get a snapshot that we can push out with a few of our applications.

brharrington:
Closing in favor of #213

PR #213, opened 2015-09-04
MetricCollector for the AWS SDK

brharrington:
Extension that allows spectator to capture data about
AWS requests being made using the SDK.

Currently this is only capturing request metrics since
it gives us the most context. For now the throughput data
from the ServiceMetricCollector is ignored.


brharrington:
Rework of #212 to address a few minor review comments and tweak the formatting a bit to be more consistent with the rest of the code.

PR #220, opened 2015-09-16
refactor TagList to be more intuitive

brharrington:
Previous behavior was based on some tests to reduce the
cost of generating the dynamic ids in high volume paths
compared to servo. However, it led to a number of surprising
and confusing issues for users. With this change we want
to clear up the behavior for users and we can optimize
as needed in the future.

It will now sort by the tag key and remove duplicates on
update. The goal is to have more intuitive behavior for
the user so they can rely on:

1. Any ids with the same set of tags will be equal.
2. Later additions will always override if the key already
   exists. So for `withTags(Iterable)` if there were
   entries with duplicate keys the iteration order would
   determine which one ended up getting used.

### DIFF
--- a/spectator-api/src/main/java/com/netflix/spectator/api/TagList.java
+++ b/spectator-api/src/main/java/com/netflix/spectator/api/TagList.java
@@ -17,15 +17,80 @@ package com.netflix.spectator.api;
 
 import com.netflix.spectator.impl.Preconditions;
 
+import java.util.Comparator;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.NoSuchElementException;
+import java.util.TreeMap;
 
 /**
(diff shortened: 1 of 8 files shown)


spectator-api/src/main/java/com/netflix/spectator/api/TagList.java
```diff
@@ -101,20 +168,115 @@ public void remove() {
...
+    }
+    return buf.toString();
+  }
+
+  /**
+   * Produces a list with with the specified tag merged with the existing values in this list.  If the key of the
+   * specified tag matches an existing list entry, then the value of the specified tag will replace the existing
+   * value.
+   *
+   * @param tag th possibly null tag to merge into the list
```
brharrington: s/th//

PR #231, opened 2015-09-25
Add support for Ids that have dynamically commuted tag values.

pstout:


spectator-api/src/main/java/com/netflix/spectator/api/DynamicId.java
```diff
@@ -0,0 +1,32 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spectator.api;
+
+/**
+ * An extension of the {@link Id} interface that allows the list of tag names to attached to the Id
```
brharrington: s/names to attached/names attached/

spectator-api/src/main/java/com/netflix/spectator/api/TagFactory.java
```diff
@@ -0,0 +1,40 @@
...
+   * produced by the createTag method.
+   */
+  String name();
+
+  /**
+   * Produces a tag based on the specified input value.
+   *
+   * @param value
+   *      possibly null input to the factory for producing a tag.  The semantics
+   *      of the are specific the TagFactory implementation.  If the value is
```
brharrington: s/of the are specific the/are specific to/

spectator-api/src/main/java/com/netflix/spectator/api/BasicTagFactory.java
```diff
@@ -0,0 +1,53 @@
...
+import com.netflix.spectator.impl.Preconditions;
+
+/**
+ * Basic implementation of the TagFactory class that does nothing fancy.
+ */
+public final class BasicTagFactory implements TagFactory {
+  private final String name;
+
+  /** Create a new instance that will produce tags with the specified non-null name. */
+  protected BasicTagFactory(String name) {
```
brharrington: protected with final class? This would be the same as default package access correct?

spectator-api/src/main/java/com/netflix/spectator/api/TagFactory.java
```diff
@@ -0,0 +1,40 @@
...
+   *      null, then an implementation may compute a value based on information
+   *      available in the runtime state of the thread.
+   * @return
+   *      the appropriate tag given the specified input value, possibly null
+   */
+  Tag createTag(String value);
```
brharrington: What is an example of how the value would get used? It looks like the DefaultDynamicId class just passes in null and it isn't clear to me where a value passed into this would come from.

spectator-api/src/main/java/com/netflix/spectator/api/DynamicId.java
pstout: Fixed.

spectator-api/src/main/java/com/netflix/spectator/api/DynamicId.java
```diff
@@ -0,0 +1,32 @@
...
+ * to be declared in advance of the use of the metric.
+ */
+public interface DynamicId extends Id {
+  /** New id with an additional tag factory.
+   * @param factory the factory to use to generate the values for the tag
+   */
+  DynamicId withTagFactory(TagFactory factory);
+
+  /** New id with additional tag factories.
+   * @param factories a map of tag name to factories for producing values for that tag
```
pstout: Relic from an earlier iteration on the code.  Fixed.

spectator-api/src/main/java/com/netflix/spectator/api/TagFactory.java
pstout: Fixed.

spectator-api/src/main/java/com/netflix/spectator/api/BasicTagFactory.java
pstout: I think this is another artifact of an earlier version of the code.  I think that the class should be final, but have a public constructor.  Thinking about this class made me realize that the ConstantTagFactory class that was in the DefaultDynamicId class should really be a standalone, public class.  I can see it being a common way of providing default values for a tag that may only be occasionally overridden when updating a metric.

The more I think about this, the more I think that the BasicTagFactory is probably not actually needed.  If someone using a DynamicId has a particular tag that they want to use they should just use the non-dynamic withTag method to specify a value.

Thoughts?

spectator-api/src/main/java/com/netflix/spectator/api/TagFactory.java
pstout: As I was writing my response to the comment on the BasicTagFactory, I started to realize that passing in a value to the createTag method was less necessary/appropriate in this world than in the DynamicCounter code.  In that world, the only way to create a Tag (aka DimensionalValue) was via the createValue method.  In the Spectator world, the withTag methods give a convenient way to create a tag and add it to an Id.

If a TagFactory implementation has non-trivial logic in it, then it should provide an appropriate method for producing a value based on whatever inputs are appropriate to that factory.  I think the createTag method should take no parameters, but be allowed to return null if it has no way of producing a meaningful value based on the input it has available in the runtime state.

Thoughts?

spectator-api/src/main/java/com/netflix/spectator/api/BasicTagFactory.java
brharrington: Agreed, it seems like the TagFactory would only need to get used if the value is unknown when creating the id.

spectator-api/src/main/java/com/netflix/spectator/api/TagFactory.java
brharrington: > I think the createTag method should take no parameters, but be allowed to return null if it has no way of producing a meaningful value based on the input it has available in the runtime state.

Sounds reasonable.

brharrington: > If a TagFactory implementation has non-trivial logic in it

One of the problems we have sometimes had with gauges is that expensive computations or operations that can get really slow like network calls would get performed inline. We consider that a bug and advise that the method should be cheap with an expectation it will be called frequently. 

Since `createTag` will get called for each increment/record I think we'll need a similar guideline. Is there some policy for this in streaming-utilities now?

spectator-api/src/main/java/com/netflix/spectator/api/BasicTagFactory.java
pstout: Removed.

spectator-api/src/main/java/com/netflix/spectator/api/Registry.java
```diff
@@ -52,6 +52,28 @@
...
+   * Creates a dynamic identifier for a meter. All ids passed into other calls should be created by the registry.
+   *
+   * @param name
+   *     Description of the measurement that is being collected.
+   * @param tagFactories
+   *     Other factories that can generate other dimensions that can be used to classify the measurement.
+   * @return
+   *     The newly created dynamic identifier.
+   */
+  DynamicId createDynamicId(String name, Iterable<TagFactory> tagFactories);
```
brharrington: Note to self, this should have caused the compatibility check in the build to fail.

spectator-api/src/main/java/com/netflix/spectator/api/TagFactory.java
pstout: Updated createTag method as discussed to remove parameter.

At least originally, the expectation with the DimensionalCounter code was that the computation of a value/tag was local to the JVM.  It might rely on previously cached data (GeoDimension and DeviceDimension for example) or do complicated pattern matching to determine the appropriate tag (DeviceDimension mapping ESN to device type), but it was local.

spectator-api/src/main/java/com/netflix/spectator/api/Registry.java
pstout: Hmm, if I don't provide a default implementation, this becomes a breaking change.  That said, I'm not sure how to provide a generic implementation of the first method.  The second method could be given a default implementation of ```return createDynamicId(name).withTagFactories(tagFactories);```.  Thoughts?

brharrington: In this case I think it is fine for it to break the api since we aren't at 1.x yet and currently all the registry implementations are local to this repo. However, I am concerned that the check in the build didn't flag the issue:

https://github.com/Netflix/spectator/blob/master/spectator-api/build.gradle#L11

I do see that it ran as part of the travis build.

brharrington:
I was trying out the dynamic ids, and I think there is another problem. The ids get resolved when read, but we need to force the resolution when the change occurs. For example:

```java
  @Test
  public void tryDynamicId() {
    Registry r = new DefaultRegistry();
    Id foo = r.createDynamicId("foo").withTagFactory(new TagFactory() {
      private int v = 0;

      @Override public String name() {
        return "bar";
      }

      @Override public Tag createTag() {
        ++v;
        final String value = "" + v;
        return new Tag() {
          @Override public String key() {
            return "bar";
          }

          @Override public String value() {
            return value;
          }
        };
      }
    });

    r.counter(foo).increment();
    print(r);
    r.counter(foo).increment();
    print(r);
    r.counter(foo).increment();
    print(r);
    print(r);
  }

  private void print(Registry r) {
    System.out.println("=======================");
    for (Meter m : r) {
      for (Measurement measurement : m.measure()) {
        System.out.println(measurement);
      }
    }
  }
```

And I get:

```
=======================
Measurement(foo:bar=1,1443627697205,1.0)
=======================
Measurement(foo:bar=2,1443627697217,2.0)
=======================
Measurement(foo:bar=3,1443627697217,3.0)
=======================
Measurement(foo:bar=4,1443627697217,3.0)
```

I was expecting to get 3 counters with a single increment.

PR #234, opened 2015-10-09
fix NPE with CompositeMeter.hasExpired

brharrington:
If an id is missing from one of the registries it was
throwing:

```
java.lang.NullPointerException
	at com.netflix.spectator.api.CompositeMeter.hasExpired(CompositeMeter.java:52)
	at com.netflix.spectator.api.CompositeRegistryTest.testHasExpired(CompositeRegistryTest.java:249)
```

@pstout this should fix the NPE bug you found.

### DIFF
--- a/spectator-api/src/test/java/com/netflix/spectator/api/CompositeRegistryTest.java
+++ b/spectator-api/src/test/java/com/netflix/spectator/api/CompositeRegistryTest.java
@@ -238,6 +238,18 @@ public class CompositeRegistryTest {
   }
 
   @Test
+  public void testHasExpired() {
+    CompositeRegistry r = new CompositeRegistry(clock);
+
+    Counter c1 = r.counter("id1");
+    Assert.assertTrue(c1.hasExpired());
+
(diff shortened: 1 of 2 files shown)


spectator-api/src/test/java/com/netflix/spectator/api/CompositeRegistryTest.java
```diff
@@ -238,6 +238,18 @@ public void testAddAndRemove() {
...
   @Test
+  public void testHasExpired() {
+    CompositeRegistry r = new CompositeRegistry(clock);
+
+    Counter c1 = r.counter("id1");
+    Assert.assertTrue(c1.hasExpired());
+
+    Registry r1 = new DefaultRegistry(clock);
+    r.add(r1);
+    Assert.assertTrue(c1.hasExpired());
```
pstout: As long as you are here you might add one more assertion that c1.hasExpired returns false after you add a counter to r1.

pstout:
LGTM

PR #246, opened 2015-11-25
add helpers for working with Streams

brharrington:
For test cases it is sometimes useful to check aggregated
values rather than individual counters. The most common
example is looking at the overall number of increments for
a counter with a given name. The recommendation was to use
streams, but that can get a bit cumbersome:

```java
long sum = StreamSupport.stream(registry.spliterator(), false)
    .filter(m -> m instanceof Counter)        // Convert to counters
    .map(m -> (Counter) m)
    .filter(c -> "foo".equals(c.id().name())) // Restrict by name
    .mapToLong(Counter::count)                // Get value
    .reduce(0L, Long::sum)                    // Compute sum
```

This change adds some helpers to make this sort of use-case
a bit easier:

```java
long sum = registry.counters()
    .filter(Functions.nameEquals("foo"))      // Restrict by name
    .mapToLong(Counter::count)                // Get value
    .reduce(0L, Long::sum)                    // Compute sum
```

There are also some examples in the unit tests and comments
for the stream methods on Registry.

### DIFF
--- a/codequality/pmd.xml
+++ b/codequality/pmd.xml
@@ -37,6 +37,7 @@
     <exclude name="DuplicateImports"/>
     <exclude name="EmptyMethodInAbstractClassShouldBeAbstract"/>
     <exclude name="ExcessiveImports"/>
+    <exclude name="ExcessivePublicCount"/>
     <exclude name="FieldDeclarationsShouldBeAtStartOfClass"/>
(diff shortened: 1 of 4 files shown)


brharrington:
@quidryan let me know what you think. Will this make your testing use-cases less cumbersome?

PR #248, opened 2015-11-25
rename to SpectatorMetricCollector

brharrington:
There is occasional confusion caused because the name
doesn't match the name for the base class in the AWS
SDK. Since there isn't much use yet, it is probably a
good time to rename to match and avoid more confusion
in the future.


brharrington:
@cfieber how much pain will this cause for you? If needed I can add in a deprecated class with the old name to avoid breaking things.

brharrington:
Based on conversation in person it should be easy to change in one-place, so we aren't going to bother with accumulating more deprecated baggage.

PR #254, opened 2016-01-14
add percentile timer and distribution summary

brharrington:
This is an alternative to using digests to get an approximate
percentile value that works well with slicing and dicing. For
this approach we are using a fixed set of buckets that partition
the range of long values so it can be used with most sets of
non-negative long values.

The error for an estimate is usually within a few percent of
the actual percentile for the real data sets tried, but there
are some exceptions. Compared with using t-Digests it is far
more performant and usually more compact. That comes at a
cost of being less accurate and more restrictive on inputs.

Docs will be updated after the cheaper bucket options are
moved from sandbox to stable. We do need to be careful with
the overhead and encourage users to be diligent about the
usage of percentiles and the overhead compared to options
such as a normal timer or bucket timer. A rule of thumb
is that a percentile timer will be ~ 300x more expensive in
terms of metrics volume than a normal timer.

### DIFF
--- a/spectator-api/src/main/java/com/netflix/spectator/api/histogram/PercentileDistributionSummary.java
+++ b/spectator-api/src/main/java/com/netflix/spectator/api/histogram/PercentileDistributionSummary.java
@@ -1,0 +1,101 @@
+/**
+ * Copyright 2015 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


spectator-api/src/main/java/com/netflix/spectator/api/histogram/PercentileDistributionSummary.java
```diff
@@ -0,0 +1,101 @@
...
+
+import com.netflix.spectator.api.Counter;
+import com.netflix.spectator.api.DistributionSummary;
+import com.netflix.spectator.api.Id;
+import com.netflix.spectator.api.Measurement;
+import com.netflix.spectator.api.Registry;
+
+import java.util.Collections;
+
+/** Timers that get updated based on the bucket for recorded values. */
```
dmuino: s/timer/distribution summary/ ?

brharrington: Yeah, looks like there are several comments there that need to get fixed. Thanks.

PR #269, opened 2016-02-19
Added support for Metrics3 Gauge

kennedyoliveira:
Hi!

I added support for Metrics3 Gauge, so when using reporting functions from Metrics3 it gets the values for Gauges too.

### DIFF
--- a/spectator-reg-metrics3/src/main/java/com/netflix/spectator/metrics3/MetricsGaugeAggr.java
+++ b/spectator-reg-metrics3/src/main/java/com/netflix/spectator/metrics3/MetricsGaugeAggr.java
@@ -1,0 +1,68 @@
+/**
+ * Copyright 2015 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spectator.metrics3;
+
+import java.util.Iterator;
+import java.util.concurrent.ConcurrentLinkedQueue;
+
+/**
+ * <p>Aggregate gauge used to compute the sum of gauges with the same Id.</p>
+ * <p>This aggregate gauge is only used to register with Metrics, the register with spectator is for each one of
+ * the registered gauges in this Aggregate gauge.</p>
+ *
+ * @author Kennedy Oliveira
+ */
+class MetricsGaugeAggr implements com.codahale.metrics.Gauge<Double> {
+
+  private ConcurrentLinkedQueue<MetricsGauge> gauges;
+
+  /**
+   * Creates a new Aggregate Gauge.
+   */
+  MetricsGaugeAggr() {
+    this.gauges = new ConcurrentLinkedQueue<>();
+  }
+
+  @Override
+  public Double getValue() {
+    double aggregatedValue = 0.0D;
+
+    final Iterator<MetricsGauge> iterator = gauges.iterator();
+    while (iterator.hasNext()) {
+      final MetricsGauge gauge = iterator.next();
+
+      if (gauge.hasExpired()) {
+        iterator.remove();
+      } else {
+        final double gaugeVal = gauge.value();
+
+        if (Double.isFinite(gaugeVal))
+          aggregatedValue += gaugeVal;
+      }
+    }
+
+    return aggregatedValue;
+  }
+
+  /**
+   * Add a gauge to be aggragate to the others.
+   *
+   * @param gauge Gauge to be added
+   */
+  void addGauge(MetricsGauge gauge) {
+    this.gauges.add(gauge);
+  }
+}
--- a/spectator-reg-metrics3/src/main/java/com/netflix/spectator/metrics3/MetricsGaugeAggr.java
+++ b/spectator-reg-metrics3/src/main/java/com/netflix/spectator/metrics3/MetricsGaugeAggr.java
@@ -38,7 +38,7 @@ class MetricsGaugeAggr implements com.codahale.metrics.Gauge<Double> {
 
   @Override
   public Double getValue() {
-    double aggregatedValue = 0.0D;
+    double aggregatedValue = Double.NaN;
 
     final Iterator<MetricsGauge> iterator = gauges.iterator();
     while (iterator.hasNext()) {
@@ -49,8 +49,10 @@ class MetricsGaugeAggr implements com.codahale.metrics.Gauge<Double> {
       } else {
         final double gaugeVal = gauge.value();
 
-        if (Double.isFinite(gaugeVal))
-          aggregatedValue += gaugeVal;
+        // When it's NaN means the gauge can be unregistered due to it's reference to the value to extract value
+        // was garbage collected or simple the gauge returned a NaN so i don't count it
+        if (!Double.isNaN(gaugeVal))
+          aggregatedValue = (Double.isNaN(aggregatedValue) ? 0 : aggregatedValue) + gaugeVal;
       }
     }
 
--- a/spectator-reg-metrics3/src/main/java/com/netflix/spectator/metrics3/MetricsRegistry.java
+++ b/spectator-reg-metrics3/src/main/java/com/netflix/spectator/metrics3/MetricsRegistry.java
@@ -17,9 +17,11 @@ package com.netflix.spectator.metrics3;
 
 import com.codahale.metrics.Gauge;
 import com.netflix.spectator.api.*;
+import com.netflix.spectator.impl.Throwables;
 
 import java.util.Collection;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.ToDoubleFunction;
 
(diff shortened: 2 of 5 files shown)


brharrington:
Thanks for the contribution. I have a lot going on today, but I should be able to review it tomorrow.

kennedyoliveira:
@brharrington, no problem, i'm glad to help!

brharrington:
Mostly looks good to me. When trying it the one problem I found was with duplicate registrations.

```java
  @Test
  public void gaugeNumberDuplicate() {
    MetricRegistry codaRegistry = new MetricRegistry();
    MetricsRegistry r = new MetricsRegistry(clock, codaRegistry);
    AtomicInteger num1 = r.gauge("foo", new AtomicInteger(42));
    AtomicInteger num2 = r.gauge("foo", new AtomicInteger(21));
    Assert.assertEquals(63.0, (Double) codaRegistry.getGauges().get("foo").getValue(), 1e-12);
  }
```

This throws an exception:

```
java.lang.IllegalArgumentException: A metric named foo already exists

	at com.codahale.metrics.MetricRegistry.register(MetricRegistry.java:91)
	at com.netflix.spectator.metrics3.MetricsGauge.<init>(MetricsGauge.java:42)
	at com.netflix.spectator.metrics3.MetricsRegistry.gauge(MetricsRegistry.java:65)
	at com.netflix.spectator.metrics3.MetricsRegistry.gauge(MetricsRegistry.java:61)
	at com.netflix.spectator.api.Registry.gauge(Registry.java:407)
	at com.netflix.spectator.metrics3.MetricsRegistryTest.gaugeNumberDuplicate(MetricsRegistryTest.java:85)
```

In general for spectator the philosophy for exceptions is to log a warning but not throw by default ([exception propagation](https://github.com/Netflix/spectator/blob/master/spectator-api/src/main/java/com/netflix/spectator/api/Throwables.java#L38)). For tests there is a property which can be set to enable throwing exceptions so it is more likely that bugs will get caught early on.

The reasoning for this is that users often don't expect or think about exceptions coming from the instrumentation code. This has caused outages at Netflix before where a precondition in the monitoring library threw an exception and caused a customer facing issue. It was a bug in the way they were using the library, but overall we prefer if the fallout of such issues is loss of visibility rather than a bigger problem.

kennedyoliveira:
@brharrington, didn't know about this philosophy, i'm new to Spectator, and i didn't make code compatible with the gauge philosophy of aggregating the metrics with the same value, i'm gonna fix it.

kennedyoliveira:
@brharrington, i got a doubt, the Throwables is a package private class, i can't use it from the code i'm implementing, there is any alternative or i'm missing something?

kennedyoliveira:
@brharrington, i refactored the code to respect the gauges contract, now it seems to work fine, i just let a FIXME to throw a exception with Throwables like you suggested but since it's package private and i can't access i'm waiting on a suggestion from you, since i'm not very familiarised with Spectator API yet.

brharrington:
@kennedyoliveira thanks. I suppose the best option is to move the Throwables utility into the [impl](https://github.com/Netflix/spectator/tree/master/spectator-api/src/main/java/com/netflix/spectator/impl) package. That is how we currently share some of the other utilities that we have to make public, but want to make clear are considered an internal api and should not be used outside of spectator itself.

spectator-reg-metrics3/src/main/java/com/netflix/spectator/metrics3/NameUtils.java
```diff
@@ -0,0 +1,51 @@
...
+ * Utility class with static methods for creating names to Metrics based on Spectator
+ * {@link com.netflix.spectator.api.Id}.
+ *
+ * @author Kennedy Oliveira
+ */
+final class NameUtils {
+
+  /**
+   * Utility Class.
+   * @deprecated
```
brharrington: Since, this constructor cannot be called (barring reflection magic), what is the benefit of marking this deprecated?

spectator-reg-metrics3/src/main/java/com/netflix/spectator/metrics3/MetricsGaugeAggr.java
```diff
@@ -0,0 +1,68 @@
...
+    final Iterator<MetricsGauge> iterator = gauges.iterator();
+    while (iterator.hasNext()) {
+      final MetricsGauge gauge = iterator.next();
+
+      if (gauge.hasExpired()) {
+        iterator.remove();
+      } else {
+        final double gaugeVal = gauge.value();
+
+        if (Double.isFinite(gaugeVal))
```
brharrington: As a note to self, the behavior for non-finite values is something that should be clarified in so the user has a clear expectation.

spectator-reg-metrics3/src/main/java/com/netflix/spectator/metrics3/MetricsGaugeAggr.java
```diff
@@ -0,0 +1,68 @@
...
+  /**
+   * Creates a new Aggregate Gauge.
+   */
+  MetricsGaugeAggr() {
+    this.gauges = new ConcurrentLinkedQueue<>();
+  }
+
+  @Override
+  public Double getValue() {
+    double aggregatedValue = 0.0D;
```
brharrington: This will result in a 0 being reported if all gauge values are expired or have non-finite values. I wonder if NaN might be more appropriate to clearly indicate that we don't know an appropriate value. Otherwise the user would probably assume that 0 was the measured value. For example, suppose we are showing a queue size, I think it would be better to indicate we don't know the queue size rather than the queue size is 0.

spectator-reg-metrics3/src/main/java/com/netflix/spectator/metrics3/MetricsRegistry.java
```diff
@@ -57,4 +54,42 @@ private String toMetricName(Id id) {
...
+
+  @Override public <T> T gauge(Id id, T obj, ToDoubleFunction<T> f) {
+    final String name = toMetricName(id);
+
+    Gauge aggrGauge = this.impl.getGauges().get(name);
+
+    // Wasn't registered already
+    if (aggrGauge == null) {
+      aggrGauge = new MetricsGaugeAggr();
+      this.impl.register(name, aggrGauge);
```
brharrington: It looks like there is a race condition here, if two threads get before the register takes place, then both will get here and one will throw.

brharrington:
I think #274, should take care of the Throwables problem you mentioned. Let me know if you still see a gap or have a better alternative.

spectator-reg-metrics3/src/main/java/com/netflix/spectator/metrics3/NameUtils.java
kennedyoliveira: I think there is no benefit, it's something i do without thinking, because i worked in a project that was a rule to make utility class constructor deprecated, i can remove since there is no point.

spectator-reg-metrics3/src/main/java/com/netflix/spectator/metrics3/MetricsGaugeAggr.java
kennedyoliveira: Thanks for the note, i guess is better to check against a NaN instead of using finite, since it will keep consistent with the gauges behaviour.

spectator-reg-metrics3/src/main/java/com/netflix/spectator/metrics3/MetricsGaugeAggr.java
kennedyoliveira: Totally agree, makes more sense and will me more consistent.

spectator-reg-metrics3/src/main/java/com/netflix/spectator/metrics3/MetricsRegistry.java
kennedyoliveira: Hmm thats true, didn't thought about that, i'm not an expert in concurrency, could you suggest a fix for that? Should i sincronize the get and register part? I don't know if it is the better alternative due to synchronisation overhead, but since this method will not be called so many times, maybe works. Anyway, i'm waiting your feedback

spectator-reg-metrics3/src/main/java/com/netflix/spectator/metrics3/NameUtils.java
brharrington: Not a big deal, I was mostly just curious. In general unless there is a user benefit to it I would bias towards leaving it out, but I don't have strong feelings on the matter.

kennedyoliveira:
@brharrington, it sure solves the problem! I rarely send pull requests so i don't know how to procede in this situation, i should wait your pr being merged to merge mine with master and get your changes or should i merge with your pr and update it already?

spectator-reg-metrics3/src/main/java/com/netflix/spectator/metrics3/MetricsRegistry.java
brharrington: I would prefer not to lock, though I agree it would it is probably a rare case. Other options:

* Catch the exception from the duplicate registration and do another get.
* Maintain our own concurrent map of registered gauges. It looks like the `getGauges` call results in building a map of all gauges so this approach might also result in less temporary garbage getting created for each registration.

brharrington:
Great! I went ahead and merged #274 so you should be able to just fetch and merge master back into your branch.

kennedyoliveira:
@brharrington, thank you! I fixed the things you mentioned, and i opted to use your  Concurrent Map suggestion for holding or registers since the metrics really build a new map each time getGauges is called, sounds like the most readable solution to me.

brharrington:
Looks good to me. Thanks again.

kennedyoliveira:
@brharrington, your welcome, i thank you for the patience haha and for the notes. I'm glad to help :D

PR #282, opened 2016-04-02
port wiki to use mkdocs

brharrington:
Initial attempt at moving the documentation from the
github wiki to mkdocs. This has the following benefits:

* Contributions to the docs can be made via PRs
* Basic search available for content of the docs


brharrington:
Sample published to: http://brharrington.github.io/spectator/

PR #317, opened 2016-09-03
add spectator-ext-placeholders subproject

brharrington:
Move the DynamicId code from the `spectator-api` project to `spectator-ext-placeholders`.

Reasons for this change:

1. Reduce confusion for users. There is some confusion
   about the difference between an Id and DynamicId. This
   is compounded by the historical context of DynamicCounter
   in servo which is more like Id than DynamicId. DynamicId has been
   renamed to PlaceholderId to better convey how it is actually used.
2. Reduce footprint of core api project. The placeholders
   can be used as a pattern on top of the core project. This
   means fewer methods in Registry and less work for Registry
   implementations.

TODO

- [x] rename classes to placeholder rather than dynamic
- [x] registry view for creating placeholder instances over an arbitrary registry
- [x] deprecate classes in spectator-api
- [x] add docs


spectator-ext-placeholders/src/main/java/com/netflix/spectator/placeholders/PlaceholderId.java
```diff
@@ -0,0 +1,72 @@
...
+ * to the Id to be declared in advance of the use of the metric.  This can be used to
+ * provide a default value for a tag or to use a TagFactory implementation that uses
+ * context available in the execution environment to compute the value of the tag.
+ */
+public interface PlaceholderId {
+  /** Description of the measurement that is being collected. */
+  String name();
+
+  /** Other dimensions that can be used to classify the measurement. */
+  Iterable<Tag> tags();
```
pstout: Given that this interface no longer extends Id, I don't think that this method should be here.  People should be using the resolveToId method instead.

spectator-ext-placeholders/src/main/java/com/netflix/spectator/placeholders/DefaultPlaceholderId.java
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  @Override
+  public int hashCode() {
+    return 31 * name.hashCode() + tagFactories.hashCode();
+  }
+
+  @Override
+  public String toString() {
+    StringBuilder buf = new StringBuilder();
```
pstout: This should be implemented as ```return resolveToId().toString();```.

spectator-ext-placeholders/src/main/java/com/netflix/spectator/placeholders/DefaultPlaceholderId.java
```diff
@@ -0,0 +1,221 @@
...
+    }
+  }
+
+  @Override
+  public DefaultPlaceholderId withTagFactories(Iterable<TagFactory> factories) {
+    return createNewId(sorter -> sorter.addFactories(factories));
+  }
+
+  @Override
+  public Id resolveToId(Registry registry) {
```
pstout: Move specifying the registry to the constructor.

spectator-ext-placeholders/src/main/java/com/netflix/spectator/placeholders/DefaultPlaceholderId.java
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  @Override
+  public String name() {
+    return name;
+  }
+
+  @Override
+  public Iterable<Tag> tags() {
+    return tagFactories.stream()
```
pstout: Move this functionality to the resolveToId method and remove this method.

Issue #312, opened 2016-08-17
Registry.{counters,timers} do not work on CompositeRegistry


brharrington:
```
scala> import com.netflix.spectator.api._
import com.netflix.spectator.api._

scala> val r = new DefaultRegistry
r: com.netflix.spectator.api.DefaultRegistry = com.netflix.spectator.api.DefaultRegistry@5121ea3c

scala> Spectator.globalRegistry.add(r)

scala> r.timer("foo").record(42, java.util.concurrent.TimeUnit.MILLISECONDS)

scala> r.counter("bar").increment()

scala> val v = r.gauge("baz", new java.util.concurrent.atomic.AtomicLong)
v: java.util.concurrent.atomic.AtomicLong = 0

scala> r.stream().forEach(m => println(m.getClass.getName))
com.netflix.spectator.api.AggrMeter
com.netflix.spectator.api.DefaultCounter
com.netflix.spectator.api.DefaultTimer
com.netflix.spectator.api.AggrMeter

scala> Spectator.globalRegistry.stream().forEach(m => println(m.getClass.getName))
com.netflix.spectator.api.CompositeMeter
com.netflix.spectator.api.CompositeMeter
com.netflix.spectator.api.CompositeMeter
com.netflix.spectator.api.CompositeMeter

scala> r.stream().filter(_.isInstanceOf[Counter]).forEach(m => println(m.getClass.getName))
com.netflix.spectator.api.DefaultCounter

scala> Spectator.globalRegistry.stream().filter(_.isInstanceOf[Counter]).forEach(m => println(m.getClass.getName))

scala> Spectator.globalRegistry.counters().forEach(m => println(m.getClass.getName))

scala> Spectator.globalRegistry.counters().forEach(m => println(m))
```

PR #318, opened 2016-09-06
Added spectator-web

ewiseblatt:
This module adds a /spectator/metrics HTTP endpoint that returns a different
JSON encoding than spring boot's /metrics endpoint. The JSON encoding contains
a richer representation of the registry to enlighten external observers.

The endpoint supports some filtering mechanisms to restrict the domain of
metrics exposed if the caller is only interested in particular metrics.

@brharrington please review.
Note the runtime is in java as everything else here, but the tests are in groovy which nothing else seems to be. I presume that is ok.

@cfieber FYI


spectator-web/src/main/java/com/netflix/spectator/controllers/filter/MeasurementFilter.java
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+
+package com.netflix.spectator.controllers.filter;
+
+import com.netflix.spectator.api.Measurement;
+import com.netflix.spectator.api.Meter;
+
+public interface MeasurementFilter {
+  public boolean keep(Meter meter, Measurement measurement);
```
brharrington: Why are both needed as part of the filter?

spectator-web/src/main/java/com/netflix/spectator/controllers/filter/X.java
```diff
@@ -0,0 +1,7 @@
+package com.netflix.spectator.controllers.filter;
+
+/**
+ * Created by ewiseblatt on 9/6/16.
+ */
+public class X {
```
brharrington: ???

spectator-web/src/main/java/com/netflix/spectator/controllers/model/ApplicationRegistry.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package com.netflix.spectator.controllers.model;
+
+import com.netflix.spectator.controllers.model.MetricValuesMap;
+
+/**
+ * An ApplicationRegistry is the encoding of the metrics from a particular application.
+ *
+ * This is only public for testing purposes so implements equals but not hash.
```
brharrington: Can these sort of classes be changed to package private. In my experience users will ignore the comments and start using public classes. Though that maybe less of a concern here as if I understand correctly it would not be used directly but by calling the web endpoint.

Why not just implement hashCode?

spectator-web/src/main/java/com/netflix/spectator/controllers/model/DataPoint.java
```diff
@@ -0,0 +1,64 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spectator.controllers.model;
+
+import com.netflix.spectator.api.Measurement;
+
+/**
+ * A serializable Measurement.
```
brharrington: I'm assuming this is by virtue of following bean conventions, it doesn't implement `java.io.Serializable`.

spectator-web/src/main/java/com/netflix/spectator/controllers/model/DataPoint.java
```diff
@@ -0,0 +1,64 @@
...
+
+  public long getT() { return timestamp; }
+  public double getV() { return value; }
+
+  public DataPoint(Long timestamp, Double value) {
+    this.timestamp = timestamp;
+    this.value = value;
+  }
+
+  public void aggregate(Measurement m) {
```
brharrington: Note to self, this aggregation is a bit odd, need to look into how it is used.

spectator-web/src/main/java/com/netflix/spectator/controllers/model/DataPoint.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+public class DataPoint {
+  static public DataPoint make(Measurement m) {
+    return new DataPoint(m.timestamp(), m.value());
+  }
+
+  public long getT() { return timestamp; }
+  public double getV() { return value; }
+
+  public DataPoint(Long timestamp, Double value) {
```
brharrington: Why are the constructor params using the wrapper classes instead of the primitive types?

spectator-web/src/main/java/com/netflix/spectator/controllers/model/MetricValuesMap.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package com.netflix.spectator.controllers.model;
+
+import java.util.HashMap;
+
+/**
+ * Maps a metric name to its values.
+ */
+public class MetricValuesMap extends HashMap<String, MetricValues> {
```
brharrington: Why not just use `HashMap<String, MetricValues>`? This class doesn't seem to add anything.

spectator-web/src/main/java/com/netflix/spectator/controllers/model/TaggedDataPoints.java
```diff
@@ -0,0 +1,79 @@
...
+
+import com.netflix.spectator.api.Id;
+import com.netflix.spectator.api.Measurement;
+import com.netflix.spectator.api.Tag;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * A collection of DataPoint instances for a comment set of TagValues.
```
brharrington: common set?

spectator-web/src/test/groovy/com/netflix/spectator/controllers/model/PrototypeMeasurementFilterSpec.groovy
```diff
@@ -0,0 +1,77 @@
...
+  }
+  Id withTags(Iterable<Tag> tags) {
+    return Id.withTags(tags);
+  }
+  Id withTags(Map<String, String> tags) {
+    return Id.withTags(tags);
+  }
+}
+
+// This is just an alias for readability in the tests.
```
brharrington: ???

spectator-web/src/main/java/com/netflix/spectator/controllers/model/TaggedDataPoints.java
```diff
@@ -0,0 +1,79 @@
...
+        tags.add(new TagValue(tag.key(), tag.value()));
+    }
+    if (aggregate) {
+      DataPoint point = null;
+      for (Measurement measurement : measurements) {
+         if (point == null) {
+           point = DataPoint.make(measurement);
+           dataPoints.add(point);
+         } else {
+           point.aggregate(measurement);
```
brharrington: If I followed this correctly, then this only gets used for AggrMeter types. However, it should have already done the aggregation before returning the measurements.

Is there a specific example input where you found this additional aggregation was needed?

spectator-web/src/main/java/com/netflix/spectator/controllers/MetricsController.java
```diff
@@ -0,0 +1,207 @@
...
+    } else if (meter instanceof Gauge) {
+      kind = "Gauge";
+    } else if ((meter instanceof Timer) || (meter instanceof LongTaskTimer)) {
+      kind = "Timer";
+    } else if (meter instanceof DistributionSummary) {
+      kind = "Distribution";
+    } else {
+      kind = meter.getClass().getName();
+      int dot = kind.lastIndexOf('.');
+      kind = kind.substring(dot + 1);
```
brharrington:
```java
kind = meter.getClass().getSimpleName();
```

spectator-web/src/main/java/com/netflix/spectator/controllers/MetricsController.java
```diff
@@ -0,0 +1,207 @@
...
+        valueList = new ArrayList<Measurement>();
+        collection.put(id, valueList);
+        new_ids.add(id);
+      }
+      valueList.add(measurement);
+    }
+    return new_ids;
+  }
+
+  public static String meterToKind(Meter meter) {
```
brharrington: I'll chat with some others on our team about this and try to get #312 fixed today. We should be able to reduce some of the pain around this.

settings.gradle
```diff
@@ -8,4 +8,5 @@ include 'spectator-api',
         'spectator-ext-sandbox',
         'spectator-ext-spark',
         'spectator-reg-metrics3',
-        'spectator-reg-servo'
+        'spectator-reg-servo',
+        'spectator-web'
```
brharrington: Since the is specific to spring, maybe `spectator-web-spring` would make more sense in case there winds up being `spectator-web-servlet` etc in the future.

brharrington:
I would rather just use junit for the tests. I'm not all that familiar with groovy and it seems like additional maintenance burden for little benefit to use a separate language and test framework for one subproject.

spectator-web/src/main/java/com/netflix/spectator/controllers/filter/X.java
ewiseblatt: Oops. Forgot to delete this. It was a placeholder class when I created the package to refactor into.

spectator-web/src/main/java/com/netflix/spectator/controllers/filter/MeasurementFilter.java
ewiseblatt: I thought filters could potentially be on meters or measurements. I'm filtering on meter names and measurement labels. Although the measurement names are kind of related, I thought one might want to filter on other meter properties such as its type (e.g. ignore distributions).

spectator-web/src/main/java/com/netflix/spectator/controllers/model/ApplicationRegistry.java
ewiseblatt: I could implement it, but then I'd need to figure out what an appropriate code is. If there's a way to do this, I'm all for it. I only implemented equals for testing.

If I make them package private classes, then I'd move them back into the controllers package, right? I'm ok with doing that. I moved them here to not muck things up. I need them for testing, but the tests can be in the same package. Should I do that?

spectator-web/src/main/java/com/netflix/spectator/controllers/model/DataPoint.java
ewiseblatt: By "serializable" I wasnt refering to java Serialization. I'm open to other ways of doing this. I was following what I'd seen elsewhere. I'd be happy explicitly serializing it rather than converting it into another object only to give to someone else to serialize, but I dont see precedence for doing that (or how one would).

spectator-web/src/main/java/com/netflix/spectator/controllers/model/DataPoint.java
ewiseblatt: I need to give a single value to stackdriver, and for that matter, report out a single value from the application. In spectator you sometimes have multiple meters for the same metric. Probably because you have independent threads and want to avoid synchronization there. This is combining that distributed implementation back into a single logical meter value.

spectator-web/src/main/java/com/netflix/spectator/controllers/model/DataPoint.java
ewiseblatt: fixed

spectator-web/src/test/groovy/com/netflix/spectator/controllers/model/PrototypeMeasurementFilterSpec.groovy
ewiseblatt: The name of the class was ridiculously long and java doesnt have aliasing so I created local classes that stripped off the outer class name to shorten them up. I can take this out if that is bad form. I'd also rename the classes but didnt have a good name.

spectator-web/src/main/java/com/netflix/spectator/controllers/model/TaggedDataPoints.java
ewiseblatt: I'm pretty sure that I've seen it though might have been other bugs in my code at the time.
I recall maybe the jvm memory used (I dont recall exact metric name). I'll investigate further when I get this hooked up and running again.

ewiseblatt:
I'll look into junit. I did groovy because that's what we're using in spinnaker where I originally wrote this. The stackdriver does a lot of mocking so will be harder but I'll move that as well.

spectator-web/src/main/java/com/netflix/spectator/controllers/MetricsController.java
ewiseblatt: thanks

spectator-web/src/main/java/com/netflix/spectator/controllers/model/TaggedDataPoints.java
ewiseblatt: This might have been a bug in my test where my test Id wasnt implementing equals correctly so AggrMeter wasnt doing the aggregation. However I do think I had seen this in spinnaker so I'm going to TODO removing this after I have a chance to put it all together and verify. My tests suggest that I dont need it, but I might not have replicated the scenario I'd seen in spinnaker.

spectator-web/src/main/java/com/netflix/spectator/controllers/model/ApplicationRegistry.java
brharrington: Package private gives more flexibility to change it later without breaking the public api.

spectator-web/src/main/java/com/netflix/spectator/controllers/model/DataPoint.java
brharrington: It's fine, just wanted to check my understanding.

ewiseblatt:
@brharrington PTAL

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/MetricsController.java
```diff
@@ -0,0 +1,229 @@
...
+      return defaultMeasurementFilter;
+    }
+
+    if (!prototypeFilterPath.isEmpty()) {
+      defaultMeasurementFilter = PrototypeMeasurementFilter.loadFromPath(
+          prototypeFilterPath);
+    } else {
+      defaultMeasurementFilter = new TagMeasurementFilter(
+          meterNameRegex, tagNameRegex, tagValueRegex);
+    }
```
brharrington: Just as an FYI, for our internal plugins that do something similar we have a single property that sets a filter expression. The expression is the query portion of the Atlas stack language: https://github.com/Netflix/atlas/wiki/Stack-Language#query. 

The main reason is that regex is hard to precisely target with dimensional ids. Also we have found that quite often filters are used to exclude so it is easier for a user to express what should not come through rather than or in addition to what should. I'm not saying use an Atlas expression and it probably isn't familiar to the users this is targeting , but I think it is worth considering how cumbersome the three regexes will be for expressing common filters.

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/filter/TagMeasurementFilter.java
```diff
@@ -0,0 +1,84 @@
...
+
+  /**
+   * Constructor.
+   */
+  public TagMeasurementFilter(String meterNameRegex, String tagNameRegex, String tagValueRegex) {
+    if (meterNameRegex != null && !meterNameRegex.isEmpty() && !meterNameRegex.equals(".*")) {
+      meterNamePattern = Pattern.compile(meterNameRegex);
+    } else {
+      meterNamePattern = null;
+    }
```
brharrington: Create a common function for checking the regex and compiling.

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/filter/TagMeasurementFilter.java
```diff
@@ -0,0 +1,84 @@
...
+        && !meterNamePattern.matcher(id.name()).matches()) {
+      return false;
+    }
+
+    if (tagNamePattern != null || tagValuePattern != null) {
+      for (Tag tag : id.tags()) {
+        boolean nameOk = tagNamePattern == null
+                         || tagNamePattern.matcher(tag.key()).matches();
+        boolean valueOk = tagValuePattern == null
+                         || tagValuePattern.matcher(tag.value()).matches();
```
brharrington: Common function for checking if pattern is null or matches.

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/filter/TagMeasurementFilter.java
```diff
@@ -0,0 +1,84 @@
...
+        if (nameOk && valueOk) {
```
brharrington: I had assumed that all tags would need to match the pattern. The comments should clarify the behavior to expect.

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/MetricsController.java
```diff
@@ -0,0 +1,229 @@
...
+  @Value("${spectator.endpoint.tagFilter.tagValueRegex:}")
+  private String tagValueRegex;
+
+  @Value("${spectator.endpoint.prototypeFilter.path:}")
+  private String prototypeFilterPath;
+
+  private MeasurementFilter defaultMeasurementFilter = null;
+
+  /**
+   * The default measurement filter is configured through properties.
```
brharrington: This should explain the precedence of the properties. How would a user know that prototype will override the regex?

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/MetricsController.java
```diff
@@ -0,0 +1,229 @@
...
+   */
+  public MeasurementFilter getDefaultMeasurementFilter() throws IOException {
```
brharrington: It looks like this is only called as part of a GET request handler, two questions:

1. Can the visibility be restricted?
2. Are there any thread safety concerns here?  I noticed it mentioned for other classes they were expected to be executed from a single thread, but wanted to ask here because I'm assuming the request handler could be called concurrently.

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/MetricsController.java
```diff
@@ -0,0 +1,229 @@
...
+  /**
+   * Internal API for encoding a registry that can be encoded as JSON.
+   * This is a helper function for the REST endpoint and to test against.
+   */
+  Map<String, MetricValues> encodeRegistry(
+        Registry sourceRegistry, MeasurementFilter filter) {
+    Map<String, MetricValues> metricMap = new HashMap<String, MetricValues>();
+
+    /**
+     * Spectator meters seam to group measurements. The meter name is the
```
brharrington: s/seam/seem/

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/MetricsController.java
```diff
@@ -0,0 +1,229 @@
...
+  private static String getWithDefault(Map<String, String> map,
+                                       String key, String defaultValue) {
+    String value = map.get(key);
+    return value == null ? defaultValue : value;
+  }
+
+  /**
+   * Collect all the meter values matching the tag pattern.
+   */
+  public static List<Id> collectValues(
```
brharrington: package private?

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/MetricsController.java
```diff
@@ -0,0 +1,229 @@
...
+      }
+      valueList.add(measurement);
+    }
+    return newIds;
+  }
+
+  /**
+   * Determine the type of a meter for reporting purposes.
+   */
+  public static String meterToKind(Meter meter) {
```
brharrington: package private?

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/MetricsController.java
ewiseblatt: I agree with your take on the three regexes (TagMeasurementFilter), though in practice, I expect this to be useful for asking for a particular meter/measurement or closely related grouping. I think the "PrototypeMeasurementFilter" would be used specify a more complete profile using include/exclude filtering to compose comprehensive viewpoints. I have not yet actually used it in a meaningful way, but I think it addresses your remarks about specifying excludes.

The atlas filter looks interesting, but I dont want to add/support a query language at this time. I think the prototype filter allows you to express similar ideas though not as concisely.

I can take out the TagMeasurementFilter and associated query parameters if you want, perhaps adding in its place an endpoint to ask for a specific metric (but I think the query parameters are more useful). I would like to add named MeasurementFilters as a query parameter so you could poll for standard viewpoints but it seems premature to add that in now.

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/MetricsController.java
```diff
@@ -0,0 +1,229 @@
...
+      }
+      Id id = measurement.id();
+
+      List<Measurement> valueList = collection.get(id);
+      if (valueList == null) {
+        valueList = new ArrayList<Measurement>();
+        collection.put(id, valueList);
+        newIds.add(id);
+      }
+      valueList.add(measurement);
```
brharrington: It doesn't look like the return type is ever used. It seems like this block could be reduced to:

```java
collection.computeIfAbsent(id, k -> new ArrayList<>()).add(measurement);
```

And change the method to return void.

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/MetricsController.java
```diff
@@ -0,0 +1,229 @@
...
+   *
+   * The result is a JSON document describing the metrics and their values.
+   * The result can be filtered using query parameters or configuring the
+   * controller instance itself.
+   */
+  @RequestMapping(method = RequestMethod.GET)
+  public ApplicationRegistry getMetrics(@RequestParam Map<String, String> filters)
+    throws IOException {
+    boolean all = filters.get("all") != null;
+    String filterMeterNameRegex = getWithDefault(filters, "meterNameRegex", "");
```
brharrington: Just use getOrDefault on the map:

```java
String filterMeterNameRegex = filters.getOrDefault("meterNameRegex", "");
```

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/filter/TagMeasurementFilter.java
ewiseblatt: Ok on the comments, but since it goes against your assumption I want to verify this makes sense. I was thinking that you'd use this to pull out all the metrics that have a particular dimension to them  (e.g. have a label with a region dimension) or are in a particular dimension (e.g. all the metrics pertaining to a particular a particular region). Matching all them would be specifying a profile of dimensions that seems more like specifying a meter itself.

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/MetricsController.java
ewiseblatt: Agree this is awkward. Is there a simple way to do this that wouldnt be ambiguous?
I dont think I should combine them (if both are specified then chain them together). The tag filters are there by default and I'd configure them to say keep everything that has labels (.+) which filters out the spring boot added metrics that we dont want in practice but seems wrong to hard code in. I guess I instead have a flag for that particular case (but question of when or how to override it remains and is more foggy)

BTW, I dont filter out the non-labeled things here, but could. Should I? I didnt want to impose that policy here unless it is a general recommendation to spectator consumers.

The profile filter is there to make it easy to build a profile for just the things you care about, especially if you are pushing metrics into another system and dont want to clog it up with stuff you dont care about. I figured if you used that mechanism, you dont need the crude tag mechanism.

On the query side, the tags seem more useful since they are just one-off usage to spot check something narrow.

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/MetricsController.java
ewiseblatt: Thread safety was a concern in the Stackdriver writer because I'm managing internal state and not worrying about it. Here I dont think it matters, does it? If there is contention then I'll just create a superfluous object, but it will get reclaimed once the reference goes away or be equivalent if it hangs around. The filters are immutable so its consistent.

Filters are part of the public interface, so having the default filter public seems reasonable.
I could reduce the visibility, but seems reasonable to allow other callers of collectValues to use it. Otherwise I guess I can add "null" semantics to mean the default, but that wouldnt get around any concurrency concerns.

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/MetricsController.java
ewiseblatt: I think this is the most useful method in the class. I use it in stackdriver to determine which values to push. If the filtering abstraction were deemed universally useful you could move filters into spectator-api and support to filter on the iterator.

This function is going one step beyond filtering, and that is to reorganize the registry by the measurement id's rather than the meter id's because I dont [think I want] the meter groupings, and it seems harder for me to consume that way.

I dont need the newId list here, but it is useful to have in my stackdriver consumer where I do some censorship and remove potentially bad metrics from the collection so it's convenient to know what's been added.

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/MetricsController.java
ewiseblatt: Ideally Meter has a public interface I could use instead.
I think I can make this package private since I dont require it. However I think others might. I'd use it in the StackdriverWriter however because of the way Stackdriver defines a counter I need to treat everything as a gauge so its intrinsic kind turns out not to matter.

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/MetricsController.java
ewiseblatt: I use it in external consumers. I could move the block elsewhere or maybe have a variant with it here in this class. What do you think?

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/filter/MeasurementFilter.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+public interface MeasurementFilter {
+  /**
+   * Meter is the source of the measurement.
+   * Measurement is the particular measurement being filtered.
+   *
+   * The concrete filter determines the criteria of what to keep.
+   * Returns true to keep, false to discard.
+   */
+  boolean keep(Meter meter, Measurement measurement);
```
brharrington: If I'm reading through the code correctly meter is never actually used and I don't think there is a good reason for filtering on the meter rather than the measurement.

I'm wondering if it might make sense to just use [Predicate<Measurement>](https://docs.oracle.com/javase/8/docs/api/java/util/function/Predicate.html) instead of a dedicated interface. This would also give chaining for free using the default `and` method on Predicate as well as likely being more familiar to users than a custom interface.

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/filter/TagMeasurementFilter.java
brharrington: It is easy to make it public later if needed, it is often quite difficult to restrict visibility later if it turns out not to be wanted in the public API. That is why I tend to bias towards minimizing the public surface area.

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/MetricsController.java
brharrington: Perhaps I'm wrong about the most frequent usage patterns of filter, but I was supposing that filters would most frequently be specified using the params on the GET request which basically maps to the TagMeasurementFilter. I probably should have placed the comment near the GET section (sorry I was jumping around a bit in the diff).

PrototypeMeasurementFilter I think addresses some of the flexibility concerns (I haven't finished going through all of that yet), but looks like it would have to get setup a priori by the app owner rather than a user querying the REST API.

> The atlas filter looks interesting, but I dont want to add/support a query language at this time.

Agreed, just mentioning it as an example.

> but I think the query parameters are more useful

Agreed, this is why I was thinking the flexibility for the tag measurement filter is all most users would have.

> I would like to add named MeasurementFilters as a query parameter so you could poll for standard viewpoints but it seems premature to add that in now.

Not sure I'm following this, do you mean something like having a config file with named filter blocks and being able to reference one by name when using the REST API?

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/filter/TagMeasurementFilter.java
brharrington: I think it makes sense for matching against a single tag key. If the tag name pattern matches multiple keys, then it becomes a bit harder to reason about IMO. That said, I don't think that improves if doing it the way I initially thought. So far I haven't thought of a use-case where matching all would make it better, it would just force going into the more ambiguous territory of matching all keys.

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/MetricsController.java
brharrington: There is already a utility function for filtering a measurement iterable:

https://github.com/Netflix/spectator/blob/master/spectator-api/src/main/java/com/netflix/spectator/api/Utils.java#L207

I'll take a look at the usage in the stackdriver PR and think about it a bit. The map bit seems a bit odd to me as in general there won't be duplicate measurement ids so each list would have a single value, but it isn't necessarily enforced right now.

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/MetricsController.java
ewiseblatt: Thanks. I looked for something like but didnt see it...

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/MetricsController.java
ewiseblatt: Removing the result for now and will worry about it later.

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/filter/MeasurementFilter.java
ewiseblatt: I could do this. While I dont actually use meter yet, I was thinking that the meter has additional attributes such as the kind that could be filtered on so if I were to add an interface, it should include that. Measurements dont know the meter they came from so I have the interface take them both..

ewiseblatt:
@brharrington This is a major overhaul that I think addresses all we've discussed so far.

spectator-web-spring/build.gradle
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+dependencies {
+  compileApi project(':spectator-api')
+
+  compile 'org.springframework.boot:spring-boot-starter-actuator:1.2.8.RELEASE'
```
brharrington: Should these use compileApi? It looks like some of the spring annotations are part of the public interface.

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/filter/MeasurementFilter.java
```diff
@@ -0,0 +1,36 @@
...
+
+import com.netflix.spectator.api.Measurement;
+import com.netflix.spectator.api.Meter;
+
+/**
+ * Interface for filtering measurements.
+ *
+ * This is used to reduce available metrics down to those of interest to a caller.
+ */
+public interface MeasurementFilter {
```
brharrington: Doesn't look like this is used anymore, I think it can be deleted.

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/MetricsController.java
```diff
@@ -0,0 +1,208 @@
...
+              .anyMatch(m -> m.id().equals(meter.id()))) {
+      kind = "Distribution";
+    } else if (meter instanceof Gauge) {
+      kind = "Gauge";
+    } else {
+      kind = meter.getClass().getSimpleName();
+    }
+    return kind;
+  }
+};
```
brharrington: s/;//

spectator-web-spring/src/test/java/com/netflix/spectator/controllers/model/TestMeter.java
```diff
@@ -0,0 +1,53 @@
...
+  public TestMeter(String name, List<Measurement> measures) {
+    myId = new TestId(name);
+    myMeasurements = measures;
+  }
+  public Id id() { return myId; }
+  public Iterable<Measurement> measure() {
+    return myMeasurements;
+  }
+  public boolean hasExpired() {
+    return expired;
```
brharrington: Looks like this is the only use of expired, it could just be:

```
return false;
```

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/MetricsController.java
```diff
@@ -0,0 +1,208 @@
...
+  @Value("${spectator.webEndpoint.prototypeFilter.path:}")
+  private String prototypeFilterPath;
+
+  private Map<Id, String> knownMeterKinds = new HashMap<Id, String>();
+  private Predicate<Measurement> defaultMeasurementFilter = null;
+
+  /**
+   * The default measurement filter is configured through properties.
+   */
+  public Predicate getDefaultMeasurementFilter() throws IOException {
```
brharrington: Shouldn't this be `Predicate<Measurement>`?

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/model/TaggedDataPoints.java
```diff
@@ -0,0 +1,125 @@
...
+  public Iterable<DataPoint> getValues() {
+    return dataPoints;
+  }
+
+  /**
+   * Constructor from a single measurement data point.
+   */
+  public TaggedDataPoints(Measurement measurement) {
+    tags = JacksonableTag.convertTags(measurement.id().tags());
+    dataPoints = Arrays.asList(DataPoint.make(measurement));
```
brharrington: Using `Collections.singletonList` here would avoid creating an unnecessary array.

spectator-web-spring/src/test/java/com/netflix/spectator/controllers/filter/PrototypeMeasurementFilterTest.java
```diff
@@ -0,0 +1,334 @@
...
+    public void testMetricToPatterns() {
+        PrototypeMeasurementFilterSpecification spec = new PrototypeMeasurementFilterSpecification();
+        spec.getInclude().put("meterA", meterSpecA);
+        spec.getInclude().put(".+B", meterSpecB);
+        spec.getInclude().put(".+C.*", meterSpecC);
+        PrototypeMeasurementFilter filter = new PrototypeMeasurementFilter(spec);
+
+        PrototypeMeasurementFilter.MeterFilterPattern meterPatternA
+            = new PrototypeMeasurementFilter.MeterFilterPattern("meterA", meterSpecA);
+        PrototypeMeasurementFilter.MeterFilterPattern meterPatternB
```
brharrington: Doesn't look like this is getting used. The middle pattern in the assert below is using an empty list.

spectator-web-spring/src/test/java/com/netflix/spectator/controllers/filter/PrototypeMeasurementFilterTest.java
```diff
@@ -0,0 +1,334 @@
...
+        Id idAYX = new TestId("counterA").withTag("tagA", "Y").withTag("tagB", "X");
+        Id idBZY = new TestId("counterB").withTag("tagA", "Z").withTag("tagB", "Y");
+
+        Assert.assertTrue(filter.test(new Measurement(idAYX, 1, 1)));
+        Assert.assertTrue(filter.test(new Measurement(idBZY, 2, 2)));
+    }
+
+    @Test
+    public void metersExcluded() {
+        PrototypeMeasurementFilterSpecification.MeterFilterSpecification meterSpec
```
brharrington: Doesn't look like this is used in the spec below.

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/MetricsController.java
```diff
@@ -0,0 +1,208 @@
...
+   * Internal API for encoding a registry that can be encoded as JSON.
+   * This is a helper function for the REST endpoint and to test against.
+   */
+  Map<String, MetricValues> encodeRegistry(
+        Registry sourceRegistry, Predicate<Measurement> filter) {
+    Map<String, MetricValues> metricMap = new HashMap<String, MetricValues>();
+
+    /**
+     * Flatten the meter measurements into a map of measurements keyed by
+     * the name and mapped to the different tag varients.
```
brharrington: s/varients/variants/

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/MetricsController.java
```diff
@@ -0,0 +1,208 @@
...
+        }
+      }
+    }
+    return metricMap;
+  }
+
+  /**
+   * Determine the type of a meter for reporting purposes.
+   *
+   * The accuracy of this method is sensiive to the registry value.
```
brharrington: s/sensiive/sensitive/

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/MetricsController.java
```diff
@@ -0,0 +1,208 @@
...
+   * we have is exactly right and spring seems to get in the way.
+   * If the registry was a bean, you might need to autowire it.
+   * But it isnt clear if all the metrics are created through that
+   * registry. The Spectator.globalRegistry, for example, iterates over
+   * all the meters using the generic iterator, but the kind-specific
+   * stream functions are empty if the registry was created as a spring
+   * bean so you need to passing in an @autowired registry.
+   *
+   * @param registry
+   *    Used to provide suplemental information (e.g. to search for the meter).
```
brharrington: s/suplemental/supplemental/

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/MetricsController.java
```diff
@@ -135,77 +128,81 @@ public ApplicationRegistry getMetrics(@RequestParam Map<String, String> filters)
...
+   *    A string such as "Counter". If the type cannot be identified as one of
+   *    the standard Spectator api interface variants, then the simple class name
+   *    is returned.
    */
-  public static String meterToKind(Meter meter) {
+  public static String meterToKind(Registry registry, Meter meter) {
     String kind;
-    if (meter instanceof Counter) {
+    if (registry.counters()
+              .anyMatch(m -> m.id().equals(meter.id()))) {
```
ewiseblatt: I'm hoping you'll add support into the interface so I can tell what type of meter I have directly. I'm only doing these iterations because it is the last resort (in practice almost everything is a CompositeMeter class so I took out checking instances).

I can cache the results in a Map<Id, String>.

I do that (Boolean) in StackdriverWriter:
  private Map<Id, Boolean> idToTimer = new HashMap<Id, Boolean>();
  public boolean meterIsTimer(Registry registry, Meter meter) {
    return idToTimer.computeIfAbsent(meter.id(), k -> {
        return registry.timers().anyMatch(m -> m.id().equals(meter.id()));
    });
  }

spectator-web-spring/build.gradle
ewiseblatt: I forgot to take that out from this branch. Spinnaker was doing it that way.

spectator-web-spring/build.gradle
```diff
@@ -15,7 +15,7 @@
  */
 
 dependencies {
-  compileApi project(':spectator-api')
+  compile project(':spectator-api')
```
brharrington: Hmm, I was actually wondering if the spring dependencies should be compileApi because the annotations from them are part of the public interface of some of the classes. I'm pretty sure the spectator-api dependency was correct as compileApi.

I don't know how familiar you are with compile vs compileApi, but the basic difference is compileApi will cause them to go into the published POM file as a compile scope dependency where compile will cause them to go into the POM as a runtime scope dependency (more background in https://github.com/nebula-plugins/nebula-publishing-plugin/issues/31).

ewiseblatt:
I addressed your comments. Didnt realize you werent notified.

PR #323, opened 2016-10-07
initial version of Atlas registry

brharrington:
An initial version of a registry that talks to Atlas
directly rather doing it via Servo. This allows apps
that do not need all of the legacy compatibility working
to shed many dependencies and simplify the usage.

This is still considered experimental for now and may
have changes to configuration and other settings.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -111,8 +111,8 @@ subprojects {
   tasks.withType(FindBugs) {
     findbugs.toolVersion "3.0.0"
     reports {
-      xml.enabled = true
-      html.enabled = false
+      xml.enabled = false
+      html.enabled = true
     }
   }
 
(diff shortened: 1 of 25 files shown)


pparth:
+1

pparth:
We needed that for a long time. How experimental is this considered?

brharrington:
We have been using it for a while. It is experimental in that we still want the option to potentially change the configuration a bit. So if you implemented the config interface there might be some minor incompatible changes. 

There are also some other changes being tested, but I don't expect them to change the existing api at all. Mostly around rollups for managing volume and some hooks for interacting with stream processing. Typically the flow already being published can be tee'd off to the stream processing systems. However, having some support locally on the client is desirable for allowing on-demand access with higher granularity either for the sampled frequency or dimensionality compared to what is published and stored all the time.

PR #327, opened 2016-10-11
Cleanup PlaceholderId Code

pstout:
Remove unused NoopPlaceholderId class.
Refactor PlaceholderFactory class into an interface and an implementation class.
Provide registry to PlaceholderIds when constructing them rather than at id resolution time.


spectator-ext-placeholders/src/main/java/com/netflix/spectator/placeholders/AbstractDefaultPlaceholderMeter.java
```diff
@@ -28,31 +27,28 @@
...
 abstract class AbstractDefaultPlaceholderMeter<T extends Meter> implements Meter {
   private final PlaceholderId id;
-  private final Registry registry;
   private final Function<Id, T> meterResolver;
 
   /**
    * Creates a new dynamic meter.
-   *
-   * @param id the dynamic id for the meter
+   *  @param id the dynamic id for the meter
```
brharrington: Seems like there is an extra space before `@param`.

PR #328, opened 2016-10-11
Add MdcTagFactory to placeholder extension.

pstout:


spectator-ext-placeholders/build.gradle
```diff
@@ -1,3 +1,8 @@
 dependencies {
   compileApi project(':spectator-api')
+
+  // The MdcTagFactoryTest requires a logging implementation that supports the MDC
+  // functionality, which the slf4j-nop (default) and slf4j-simple implementations do not.
```
brharrington: What is the behavior if a user is running with an implementation that doesn't support MDC?

spectator-ext-placeholders/src/main/java/com/netflix/spectator/placeholders/MdcTagFactory.java
```diff
@@ -0,0 +1,55 @@
...
+
+import com.netflix.spectator.api.BasicTag;
+import com.netflix.spectator.api.Tag;
+import com.netflix.spectator.impl.Preconditions;
+import org.slf4j.MDC;
+
+/**
+ * A TagFactory implementation that extracts information from the Sl4fj MDC data.
+ * If the MDC for the current thread has no value associated with the specified
+ * naem at the time that the createTag method is invoked, then that method will
```
brharrington: s/naem/name/

spectator-ext-placeholders/build.gradle
pstout: You get no tags from the MdcTagFactory as the underlying `MDC.get()` call always returns null.

spectator-ext-placeholders/src/main/java/com/netflix/spectator/placeholders/MdcTagFactory.java
pstout: Fixed.

PR #335, opened 2016-10-20
Softened spring dependencies

ewiseblatt:
Fewer dependencies, but still requires spring auto configure so that the
endpoint bean can be disabled by default.

Uses compileOnly to require application using this module to supply concrete
versions.

@cfieber, @ttomsu
This is intended to complement the PR I'm sending to kork. It works for me, but is it correct?

### DIFF
--- a/spectator-web-spring/build.gradle
+++ b/spectator-web-spring/build.gradle
@@ -17,7 +17,9 @@
 dependencies {
   compile project(':spectator-api')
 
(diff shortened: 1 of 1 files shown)


ewiseblatt:
@cfieber can you have a look so I can ensure this gets into Brian's release?

brharrington:
Closing in favor of #340.

PR #344, opened 2016-11-06
avoid overloaded methods with different SAM types

brharrington:
Deprecates the `record(Runnable)` and `record(Callable)`
on timers and replaces them with `run(Runnable)` and
`call(Callable)` respectively. This avoids issues with
ambiuous overloaded definitions such as:

```
scala> t.call { () => foo }
bar

scala> t.run { () => foo }
bar

scala> t.record { () => foo }
<console>:17: error: ambiguous reference to overloaded definition,
both method record in trait Timer of type (x$1: Runnable)Unit
and  method record in trait Timer of type [T](x$1: java.util.concurrent.Callable[T])T
match argument types (() => Unit)
       t.record { () => foo }
```

### DIFF
--- a/docs/intro/clock.md
+++ b/docs/intro/clock.md
@@ -72,7 +72,7 @@ ManualClock clock = new ManualClock();
 Registry registry = new DefaultRegistry(clock);
 
 Timer timer = registry.timer("test");
-timer.record(() -> {
+timer.run(() -> {
   doSomething();
   clock.setMonotonicTime(42L);
 });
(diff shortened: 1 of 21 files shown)


brharrington:
On hold for now, another area that would be nice to have some improvement on is exceptions with the java 8 lambdas. `Callable` throws `Exception` and so it forces that on the user and cannot be used with lambdas that have void return type. `Runnable` does not throw, so is cumbersome to use if the inner code throws.

PR #346, opened 2016-11-08
Add bitbucket.conf to capture Atlassian Bitbucket Server JMX metrics

ebukoski:


spectator-agent/src/main/resources/bitbucket.conf
```diff
@@ -0,0 +1,132 @@
+// http://confluence.atlassian.com/bitbucketserver/enabling-jmx-counters-for-performance-monitoring-776640189.html
+netflix.spectator.agent.jmx {
+  mappings = ${?netflix.spectator.agent.jmx.mappings} [
+    {
+      query = "com.atlassian.bitbucket.thread-pools:name=*"
+      measurements = [
+        {
+          name = "bitbucket.{name}.MaximumPoolSize"
```
brharrington: What is `name` in this example? I'm assuming the name of the thread pool which could be dynamic, in which case it should probably get moved to be a tag. The name and tag keys on a metric id should be deterministic and not contain any dynamic elements. The only dynamic part should be the values for some tags.

spectator-agent/src/main/resources/bitbucket.conf
```diff
@@ -0,0 +1,132 @@
...
+          value = "{MaximumPoolSize}"
+          tags = [ ]
```
brharrington: This should set `atlas.dstype=gauge`. See example:

https://github.com/Netflix/spectator/blob/master/spectator-agent/src/main/resources/cassandra.conf#L1533

And for background:

https://github.com/Netflix/atlas/wiki/Concepts#normalization

spectator-agent/src/main/resources/bitbucket.conf
```diff
@@ -0,0 +1,132 @@
...
+          value = "{QueueLength}"
+          tags = [ ]
+        }
+      ]
+    },
+    {
+      query = "com.atlassian.bitbucket:name=ScmStatistics"
+      measurements = [
+        {
+          name = "bitbucket.scm.Pulls"
```
brharrington: Per the docs Pulls/Pushes are monotonic counter since the process was started. They should use `atlas.dstype=rate` with specify the counter config option, e.g.:

https://github.com/Netflix/spectator/blob/master/spectator-agent/src/main/resources/cassandra.conf#L1579

spectator-agent/src/main/resources/bitbucket.conf
```diff
@@ -0,0 +1,132 @@
...
+    {
+      query = "com.atlassian.bitbucket.thread-pools:name=*"
+      measurements = [
+        {
+          name = "bitbucket.{name}.MaximumPoolSize"
+          value = "{MaximumPoolSize}"
+          tags = [ ]
+        },
+        {
+          name = "bitbucket.{name}.PoolSize"
```
brharrington: It is more idiomatic to use lower camel case, i.e., `s/PoolSize/poolSize/`.

spectator-agent/src/main/resources/bitbucket.conf
```diff
@@ -5,126 +5,285 @@ netflix.spectator.agent.jmx {
       query = "com.atlassian.bitbucket.thread-pools:name=*"
       measurements = [
         {
-          name = "bitbucket.{name}.MaximumPoolSize"
+          name = "bitbucket.threadpool.maximumPoolSize"
           value = "{MaximumPoolSize}"
-          tags = [ ]
+          tags = [
+            {
+              key = "type"
```
brharrington: Due to some unfortunate historical baggage with `type` I would recommend avoiding it. In particular servo will automatically overwrite that tag. I don't think it will matter for this specific use-case, but just to avoid any problems. Many things use `id` as the generic key for this sort of thing, e.g.:

http://netflix.github.io/spectator/en/latest/ext/jvm-memory-pools/#metric-dimensions

ebukoski: I'll change it to "pooltype"

PR #348, opened 2016-11-11
fix #347, overhead of creating PercentileTimers

### LINKED ISSUE #347: overhead of creating new PercentileTimers
In some apps there are reports of a high overhead for creating PercentileTimers due to string allocations for the tag. Need to profile this some more.

brharrington:
This change reduces the overhead of creating new instances
of `PercentileTimer` and `PercentileDistributionSummary`.
The initial state:

```
Benchmark                   Mode  Cnt         Score        Error  Units
defaultTimerGet            thrpt   50  23076956.298 ± 416682.233  ops/s
percentileTimerGet         thrpt   50      4720.225 ±     81.657  ops/s
```

Looking at where the time was being spent, most of it was
for the `String.format` calls which uses regex internally:

```
 28.8%  28.8% java.util.regex.Pattern$BmpCharProperty.match
 21.1%  21.1% java.util.Formatter.parse
 14.3%  14.3% java.lang.String.toUpperCase
  7.2%   7.3% com.netflix.spectator.api.AbstractRegistry.counter
  6.9%   6.9% com.netflix.spectator.api.histogram.PercentileTimer.<init>
  6.1%   6.1% java.util.Formatter$FormatSpecifier.print
  3.8%   3.8% com.netflix.spectator.perf.generated.PercentileTimers_percentileTimerGet_jmhTest.percentileTimerGet_thrpt_jmhStub
  3.1%   3.1% java.util.regex.Pattern$GroupHead.match
  2.3%   2.3% java.util.Formatter.format
  1.1%   1.1% java.util.regex.Pattern$Branch.match
  5.1%   5.1% <other>
```

Changing this to use precomputed string values gave a
10x improvement:

```
Benchmark                   Mode  Cnt         Score        Error  Units
defaultTimerGet            thrpt   50  23333624.790 ± 254811.495  ops/s
percentileTimerGet         thrpt   50     49980.978 ±    368.642  ops/s
```

The highest overhead after that was for prefetching the
counter values for all buckets. The prefetching was done
to avoid a lookup on the record call when a timer instance
is stored locally and reused. Though looking up in a
concurrent map is not a high overhead, it is far more than
accessing a local variable. Overhead of reusing a timer
instance:

```
Benchmark                     Mode  Cnt         Score        Error  Units
defaultTimerReuse            thrpt   50  76899727.142 ± 909947.426  ops/s
percentileTimerReuse         thrpt   50  49408513.274 ± 527442.569  ops/s
```

To further reduce the overhead of single use timers where
the lookup is preformed each time we now lazily fill in
the array of counters for each bucket. This does not
materially impact the more import reuse case:

```
Benchmark                                      Mode  Cnt         Score         Error  Units
PercentileTimers.defaultTimerGet              thrpt   50  23104705.505 ±  266616.607  ops/s
PercentileTimers.defaultTimerReuse            thrpt   50  74964025.988 ± 2652454.482  ops/s
PercentileTimers.percentileTimerGet           thrpt   50   3484868.427 ±   99621.697  ops/s
PercentileTimers.percentileTimerReuse         thrpt   50  48163827.543 ±  551056.197  ops/s
```

Overall, these changes improve the `PercentileTimer.get`
performance by about 738x.


brharrington:
fyi @dmuino @jhspaybar

spectator-api/src/main/java/com/netflix/spectator/api/histogram/PercentileDistributionSummary.java
```diff
@@ -70,10 +83,23 @@ public static PercentileDistributionSummary get(Registry registry, Id id) {
     return summary.hasExpired();
   }
 
+  // Lazily load the counter for a given bucket. This avoids the allocation for
+  // creating the id and the map lookup after the first time a given bucket is
+  // accessed for a timer.
+  private Counter counterFor(int i) {
+    Counter c = counters.get(i);
+    if (c == null) {
```
jhspaybar: Do you need to be doing some putIfAbsent type stuff through here?

PR #350, opened 2016-11-14
Added startTime to metric response.

ewiseblatt:
@brharrington. @cfieber

### DIFF
--- a/spectator-web-spring/src/main/java/com/netflix/spectator/controllers/model/ApplicationRegistry.java
+++ b/spectator-web-spring/src/main/java/com/netflix/spectator/controllers/model/ApplicationRegistry.java
@@ -16,6 +16,7 @@
 
 package com.netflix.spectator.controllers.model;
 
+import java.lang.management.ManagementFactory;
 import java.util.Map;
 import java.util.Objects;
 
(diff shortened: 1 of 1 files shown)


spectator-web-spring/src/main/java/com/netflix/spectator/controllers/model/ApplicationRegistry.java
```diff
@@ -27,14 +29,15 @@
  * This is only public for testing purposes so implements equals but not hash.
  */
 public class ApplicationRegistry {
+  static private long startTime = ManagementFactory.getRuntimeMXBean().getStartTime();
```
cfieber: `private static` (modifier ordering)

spectator-web-spring/src/main/java/com/netflix/spectator/controllers/model/ApplicationRegistry.java
```diff
@@ -16,6 +16,8 @@
 
 package com.netflix.spectator.controllers.model;
 
+import java.lang.management.ManagementFactory;
+import java.lang.management.RuntimeMXBean;
```
cfieber: unused

ewiseblatt:
fixed

PR #351, opened 2016-11-29
add constructor parameter for ConcurrentMap, default to larger Concur…

tcellucci:
…rentHashMap

### DIFF
--- a/spectator-api/src/main/java/com/netflix/spectator/api/AbstractRegistry.java
+++ b/spectator-api/src/main/java/com/netflix/spectator/api/AbstractRegistry.java
@@ -20,6 +20,7 @@ import com.netflix.spectator.impl.Preconditions;
 
 import java.util.Iterator;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 
 /**
  * Base class to make it easier to implement a simple registry that only needs to customise the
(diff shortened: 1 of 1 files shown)


spectator-api/src/main/java/com/netflix/spectator/api/AbstractRegistry.java
```diff
@@ -51,9 +52,23 @@ public AbstractRegistry(Clock clock) {
    *     Configuration settings for the registry.
    */
   public AbstractRegistry(Clock clock, RegistryConfig config) {
+    this(clock, config, new ConcurrentHashMap<>(1024));
```
brharrington: This is setting the initial size not the concurrency level. Can you create a [jmh test](https://github.com/Netflix/spectator/tree/master/spectator-api/src/jmh/java/com/netflix/spectator/perf) showing the impact?

tcellucci: the concurrency level is derived from the initial size with this particular constructor:
        int cap = ((initialCapacity >= (MAXIMUM_CAPACITY >>> 1)) ?
                   MAXIMUM_CAPACITY :
                   tableSizeFor(initialCapacity + (initialCapacity >>> 1) + 1));

tcellucci:
didn't pan out, closing

PR #356, opened 2016-12-13
Update bitbucket.conf with correct atlas.dstype

ebukoski:
"count" and "counter" are NOT valid atlas.dstypes

### DIFF
--- a/spectator-agent/src/main/resources/bitbucket.conf
+++ b/spectator-agent/src/main/resources/bitbucket.conf
@@ -71,7 +71,7 @@ netflix.spectator.agent.jmx {
           tags = [
             {
               key = "atlas.dstype"
-              value = "counter"
+              value = "rate"
             }
           ]
         },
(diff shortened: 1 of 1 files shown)


spectator-agent/src/main/resources/bitbucket.conf
```diff
@@ -71,7 +71,7 @@ netflix.spectator.agent.jmx {
           tags = [
             {
               key = "atlas.dstype"
-              value = "counter"
+              value = "rate"
```
brharrington: According to the [docs](https://confluence.atlassian.com/bitbucketserver/enabling-jmx-counters-for-performance-monitoring-776640189.html) pushes and pulls are:

> Number of scm pulls serviced by this instance since it was started

In which case you would want to add `counter = true`:

<pre>
        {
          name = "bitbucket.scm.pulls"
          value = "{Pulls}"
          <b>counter = true</b>
          tags = [
            {
              key = "atlas.dstype"
              value = "rate"
            }
          ]
        },
</pre>

PR #363, opened 2017-01-23
Extend visibility of MeasurementSerializer and PublishPayload

krutsko:
I'm using AtlasRegistry, but have a different http client for publishing to atlas.
So, I would like to reuse some basic components.


pparth:
+1

brharrington:
I will have to think about this some more. We really want to keep the about the public API minimal as it is hard to undo later. In particular, we don't want jackson exposed as part of the public API as there is some desire to shadow it so we don't create any issues for consumers that may require different versions of that jackson.

brharrington:
Closing in favor of #386.

PR #370, opened 2017-02-09
A counter that keeps track of its last update

### LINKED ISSUE #369: support for counter pattern with time since last update
A pattern that is sometimes useful internally is a counter that in addition to the count tracks the time since the last update. This is used in some circumstances where there may be low activity for a period of time and there is a desire to alert on longer gaps.

Counter would have two stats:

* `count`: number of events per second
* `interval`: represents the time since the last increment, similar to `duration` for long task timers

/cc @dmuino @ken-chu

dmuino:
This adds ActivityCounter: a counter that keeps track of the last time
it was incremented. This fixes #369

Currently this is added to spectator-api as a first approach. It follows
the same convention of the Percentile meters where the factory method
takes a Registry as a parameter to avoid polluting the registry with
extra methods.


spectator-api/src/main/java/com/netflix/spectator/api/ActivityCounter.java
```diff
@@ -0,0 +1,65 @@
+package com.netflix.spectator.api;
```
brharrington: Missing license header.

spectator-api/src/main/java/com/netflix/spectator/api/ActivityCounter.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.util.Collections;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * A counter that also keeps track of the time since last update.
+ */
+public final class ActivityCounter implements Counter {
```
brharrington: I wonder if something like `IntervalCounter` would be better at conveying how this is different. Arguably all counters track the amount of activity.

spectator-api/src/main/java/com/netflix/spectator/api/ActivityCounter.java
```diff
@@ -0,0 +1,65 @@
...
+
+  /** Create a new instance. */
+  public static ActivityCounter get(Registry registry, Id id) {
+    return new ActivityCounter(registry, id);
+  }
+
+  private static final Tag INTERVAL_TAG = new BasicTag("statistic", "interval");
```
brharrington: Do you think we should add this to the `Statistic` enum?

spectator-api/src/main/java/com/netflix/spectator/api/ActivityCounter.java
```diff
@@ -0,0 +1,65 @@
...
+
+  /**
+   * Create a new ActivityCounter using the given registry and base id.
+   */
+  ActivityCounter(Registry registry, Id id) {
+    this.registry = registry;
+    this.id = id;
+    this.counter = registry.counter(id.withTag(Statistic.count));
+    this.lastUpdated = new AtomicLong(0L);
+    registry.gauge(id.withTag(INTERVAL_TAG), lastUpdated, Functions.age(registry.clock()));
```
brharrington: I'm not sure this will work as the user would expect. If they do not keep a strong reference to the activity counter then the gauge will go away. I could see users expecting to do:

```java
Id id = ...
ActivityCounter.get(registry, id).increment();
```

dmuino:
Will refactor using the new `state()` functionality of the registry.

PR #441, opened 2017-08-03
add helper for adding boolean tag values

### LINKED ISSUE #439: withTag(String, boolean) method
As a user, I would like to be able to call withTag with a boolean value. Currently withTag only accepts a String. 

I understand the argument of spectator not getting into the business of having opinions of how to serialize objects to Strings, but I would assert that booleans are special:
* Boolean::toString defines a canonical formatting of a boolean in the JDK
* There is a established cardinality of 2 for a boolean, so there is no ability for a caller to abuse spectator supporting a type, unlike if Spectator support a withTag that accepted Long or Integer.
* It's an extremely common case. I would say 50% of my with withTag calls look like: `id.withTag("isRelevant", Boolean.toString(isRelevant))`

brharrington:
Fixes #439. See issue for details.

### DIFF
--- a/spectator-api/src/main/java/com/netflix/spectator/api/Id.java
+++ b/spectator-api/src/main/java/com/netflix/spectator/api/Id.java
@@ -34,6 +34,19 @@ public interface Id {
   Id withTag(Tag t);
 
   /**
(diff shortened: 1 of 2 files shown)


quidryan:
batman-dancing

Yeah!

PR #459, opened 2017-08-22
move polled gauges to patterns package

brharrington:
The polled gauges are just a pattern for something that is
regularly polled in the background and updates an active
gauge type on the registry. One common source of confusion
is the registry class has gauge overloads for both the
active and passive variants. With this change the passive
variants have been deprecated on the Registry interface
and the PolledGauge class can be used to support the
passive use-cases.


spectator-api/src/main/java/com/netflix/spectator/api/patterns/MonotonicCounter.java
```diff
@@ -32,72 +32,135 @@
...
+        Tuple<T> t = (Tuple<T>) c;
+        t.schedule(registry, executor);
+      }
+
+      return obj;
     }
-    return value;
   }
 
   /** Keep track of the object reference, counter, and other associated bookkeeping info. */
```
zimmermatt: `Tuple` does not connote the semantics of this type to me.  Consider something like `Bookkeeper`, `DeltaCalculator`, or something else more descriptive.

spectator-api/src/main/java/com/netflix/spectator/api/patterns/PolledGauge.java
```diff
@@ -0,0 +1,402 @@
...
+ * accurate tracking such as max and average number of connections across an interval of time.
+ * The polled gauge would not provide that level of detail.</p>
+ *
+ * <p>If multiple values are monitored with the same id, then the values will be aggregated and
+ * the sum will be reported. For example, registering multiple gauges for active threads in
+ * a thread pool with the same id would produce a value that is the overall number
+ * of active threads. For other behaviors, manage it on the user side and avoid multiple
+ * registrations.</p>
+ */
+public final class PolledGauge {
```
zimmermatt: This class and `MonotonicCounter` have a lot of commonality resulting in significant code and documentation duplication. Consider distilling out the common bits.

spectator-api/src/main/java/com/netflix/spectator/api/patterns/PolledGauge.java
```diff
@@ -0,0 +1,402 @@
...
+      Utils.propagateTypeError(registry, meter.id(), MeterTuple.class, c.getClass());
+    } else {
+      MeterTuple t = (MeterTuple) c;
+      t.add(meter);
+      t.schedule(registry, null);
+    }
+  }
+
+  /** Base class for gauge tuples used for bookkeeping. */
+  abstract static class GaugeTuple {
```
zimmermatt: Similar comment as on `Tuple` for this type and its sub-types.

spectator-api/src/main/java/com/netflix/spectator/api/patterns/PolledGauge.java
```diff
@@ -0,0 +1,402 @@
...
+      }
+      gauge.set(sum);
+    }
+  }
+
+  /**
+   * Pair consisting of weak reference to an object and a function to sample a numeric
+   * value from the object.
+   */
+  static final class RefPair<T> {
```
zimmermatt: Consider a more expressive name such as `BoundFunction`, `AppliedFunction`, or `WeakRefAppliedFunction`.

spectator-api/src/main/java/com/netflix/spectator/api/patterns/PolledGauge.java
zimmermatt: I like the fluent style.  The names, as is, are fine.  However, I'm offering up some alternatives to consider that may have a more expressive feel (see below).

spectator-api/src/main/java/com/netflix/spectator/api/patterns/PolledGauge.java
```diff
@@ -0,0 +1,402 @@
...
+  /**
+   * Return a builder for configuring a polled gauge reporting to the provided registry.
+   *
+   * @param registry
+   *     Registry that will maintain the state and receive the sampled values for the
+   *     configured gauge.
+   * @return
+   *     Builder for configuring a polled gauge.
+   */
+  public static IdBuilder<Builder> using(Registry registry) {
```
zimmermatt: Consider `registeringWith`

spectator-api/src/main/java/com/netflix/spectator/api/patterns/PolledGauge.java
```diff
@@ -0,0 +1,402 @@
...
+  }
+
+  /**
+   * Explicitly disable polling for the gauge registered with {@code id}. This is optional
+   * and is mostly used if it is desirable for the gauge to go away immediately. The polling
+   * will stop automatically when the referred object is garbage collected. See
+   * {@link Builder#monitor(Object, ToDoubleFunction)} for more information.
+   */
+  public static void remove(Id id) {
+
```
zimmermatt: Will this be implemented in a later commit?

spectator-api/src/main/java/com/netflix/spectator/api/patterns/PolledGauge.java
```diff
@@ -0,0 +1,402 @@
...
+    /**
+     * Set the executor to be used for polling the value. If not set, then the default
+     * executor will be used which is limited to a single thread to minimize the worst
+     * case resource usage for collecting the gauge data. Use a custom executor if more
+     * resources are needed or if the polling operation is expensive.
+     *
+     * @return
+     *     This builder instance to allow chaining of operations.
+     */
+    public Builder withExecutor(ScheduledExecutorService executor) {
```
zimmermatt: Consider `schedulingWith` or `schedulingExecutionWith`.

spectator-api/src/main/java/com/netflix/spectator/api/patterns/PolledGauge.java
```diff
@@ -0,0 +1,402 @@
...
+     * {@link java.util.concurrent.atomic.AtomicInteger}
+     * or {@link java.util.concurrent.atomic.AtomicLong}.
+     *
+     * @param number
+     *     Thread-safe implementation of {@link Number} used to access the value.
+     * @return
+     *     The number that was passed in to allow the builder to be used inline as part
+     *     of an assignment.
+     */
+    public <T extends Number> T monitor(T number) {
```
zimmermatt: Consider `monitorValue` or `monitoringValue`.

spectator-api/src/main/java/com/netflix/spectator/api/patterns/PolledGauge.java
```diff
@@ -0,0 +1,402 @@
...
+     * @param obj
+     *     Object used to compute a value.
+     * @param f
+     *     Function that is applied on the value for the number.
+     * @return
+     *     The object that was passed in so the registration can be done as part of an assignment
+     *     statement.
+     */
+    @SuppressWarnings("unchecked")
+    public <T> T monitor(T obj, ToDoubleFunction<T> f) {
```
zimmermatt: Consider `monitorValue` or `monitoringValue`.

spectator-api/src/main/java/com/netflix/spectator/api/patterns/LongTaskTimer.java
```diff
@@ -68,8 +68,14 @@ public static LongTaskTimer get(Registry registry, Id id) {
   private LongTaskTimer(Registry registry, Id id) {
     this.clock = registry.clock();
     this.id = id;
-    registry.monitorValue(id.withTag(Statistic.activeTasks), this, LongTaskTimer::activeTasks);
-    registry.monitorValue(id.withTag(Statistic.duration),    this, t -> t.duration() / NANOS_PER_SECOND);
+    PolledGauge.using(registry)
+        .withId(id)
+        .withTag(Statistic.activeTasks)
+        .monitor(this, LongTaskTimer::activeTasks);
```
brharrington: The fields have already been initialized when this is called so it shouldn't be a problem. In this case they can be moved after the object is constructed though as the construction is limited to a particular method.

spectator-api/src/main/java/com/netflix/spectator/api/patterns/PolledGauge.java
brharrington: I went with `scheduleOn` which is similar to the `observeOn` for rxjava Flowable.

brharrington:
Re the checks, a lot of the more common ones are already covered by the id. I'll add some additional checks in a separate PR and map them to [Registry.propagate](http://netflix.github.io/spectator/en/latest/javadoc/spectator-api/com/netflix/spectator/api/Registry.html#propagate-java.lang.String-java.lang.Throwable-). Right now though I mainly want to get the changes to the public API settled and then merge so the snapshot can be tried with a few larger code bases before becoming final.

Issue #419, opened 2017-06-29
Add utility function to monitor thread pools


ajoymajumdar:
Servo had a helper function to register pools for monitoring.

Ex:  final CompositeMonitor<?> newThreadPoolMonitor =
            Monitors.newThreadPoolMonitor("event.pool", (ThreadPoolExecutor) executor);

Add an equivalent of the above in spectator.

brharrington:
Fixed in #419.

PR #463, opened 2017-08-26
Issue #419 Add thread pool monitor.

zimmermatt:
Added polled monitoring of common thread pool properties.  Where
possible, Tomcat thread pool and executor metric names were used for
consistency.

### DIFF
--- a/spectator-api/src/main/java/com/netflix/spectator/api/patterns/ThreadPoolMonitor.java
+++ b/spectator-api/src/main/java/com/netflix/spectator/api/patterns/ThreadPoolMonitor.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2014-2017 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package com.netflix.spectator.api.patterns;
 
 import com.netflix.spectator.api.BasicTag;
@@ -49,37 +64,37 @@ public final class ThreadPoolMonitor {
   /**
    * Task count meter name.
    */
-  public static final String TASK_COUNT = "threadpool.taskCount";
+  static final String TASK_COUNT = "threadpool.taskCount";
 
   /**
    * Completed task count meter name.
    */
-  public static final String COMPLETED_TASK_COUNT = "threadpool.completedTaskCount";
+  static final String COMPLETED_TASK_COUNT = "threadpool.completedTaskCount";
 
   /**
    * Current threads busy meter name.
    */
-  public static final String CURRENT_THREADS_BUSY = "threadpool.currentThreadsBusy";
+  static final String CURRENT_THREADS_BUSY = "threadpool.currentThreadsBusy";
 
   /**
    * Max threads meter name.
    */
-  public static final String MAX_THREADS = "threadpool.maxThreads";
+  static final String MAX_THREADS = "threadpool.maxThreads";
 
   /**
    * Pool size meter name.
    */
-  public static final String POOL_SIZE = "threadpool.poolSize";
+  static final String POOL_SIZE = "threadpool.poolSize";
 
   /**
    * Core pool size meter name.
    */
-  public static final String CORE_POOL_SIZE = "threadpool.corePoolSize";
+  static final String CORE_POOL_SIZE = "threadpool.corePoolSize";
 
   /**
    * Queue size meter name.
    */
-  public static final String QUEUE_SIZE = "threadpool.queueSize";
+  static final String QUEUE_SIZE = "threadpool.queueSize";
 
   // prevent direct instantiation.
   private ThreadPoolMonitor() { }
--- a/spectator-api/src/main/java/com/netflix/spectator/api/patterns/ThreadPoolMonitor.java
+++ b/spectator-api/src/main/java/com/netflix/spectator/api/patterns/ThreadPoolMonitor.java
@@ -103,10 +103,10 @@ public final class ThreadPoolMonitor {
    * Register the provided thread pool, optionally tagged with a name.
    *
    * @param registry the registry to use
-   * @param threadPool the thread pool to monitor
+   * @param threadPool the thread pool on which to attach monitoring
    * @param threadPoolName a name with which to tag the metrics or {@code null} for the default name
    */
-  public static void monitor(
+  public static void attach(
       final Registry registry,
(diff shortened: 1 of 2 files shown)


spectator-api/src/test/java/com/netflix/spectator/api/patterns/ThreadPoolMonitorTest.java
```diff
@@ -0,0 +1,280 @@
...
+  @Test
+  public void taskCountUpdates() throws InterruptedException {
+    Meter meter = getMeter(ThreadPoolMonitor.TASK_COUNT);
+    Iterator<Measurement> measurements = meter.measure().iterator();
+    assertEquals(0.0, measurements.next().value(), 0.0);
+
+    Cancellable command = new Cancellable();
+    threadPoolExecutor.execute(command);
+
+    Thread.sleep(60);
```
brharrington: I would prefer to avoid the use of sleep in tests. In my experience that tends to make them extremely brittle. A latch should work here:

```java
final CountDownLatch latch = new CountDownLatch(1);
threadPoolExecutor.execute(latch::countDown);
latch.await(5, TimeUnit.SECONDS);
```

spectator-api/src/test/java/com/netflix/spectator/api/patterns/ThreadPoolMonitorTest.java
```diff
@@ -0,0 +1,280 @@
...
+  @Test
+  public void currentThreadsBusyCountUpdates() throws InterruptedException {
+    Meter meter = getMeter(ThreadPoolMonitor.CURRENT_THREADS_BUSY);
+    Iterator<Measurement> measurements = meter.measure().iterator();
+    assertEquals(0.0, measurements.next().value(), 0.0);
+
+    Cancellable command = new Cancellable();
+    threadPoolExecutor.execute(command);
+
+    Thread.sleep(60);
```
brharrington: Avoid sleep, perhaps something like:

```java
    final CountDownLatch started = new CountDownLatch(1);
    final CountDownLatch okToComplete = new CountDownLatch(1);
    final CountDownLatch completed = new CountDownLatch(1);

    final ThreadPoolExecutor threadPoolExecutor =
        new ThreadPoolExecutor(3, 10, 0L, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<>()) {
          @Override protected void afterExecute(Runnable r, Throwable t) {
            completed.countDown();
          }
        };

    threadPoolExecutor.execute(() -> {
      started.countDown();
      try {
        okToComplete.await();
      } catch (InterruptedException e) {
        Thread.currentThread().interrupt();
      }
    });

    started.await();
    System.out.println(threadPoolExecutor.getActiveCount());

    okToComplete.countDown();
    completed.await();
    System.out.println(threadPoolExecutor.getActiveCount());
    
    threadPoolExecutor.shutdownNow();
```

spectator-api/src/main/java/com/netflix/spectator/api/patterns/ThreadPoolMonitor.java
```diff
@@ -0,0 +1,133 @@
...
+ * Monitors and reports a common set of metrics for thread pools.
+ *
+ * <p>The following is the list of metrics reported followed by which {@link ThreadPoolExecutor}
+ * method populates it.</p>
+ *
+ * <pre>
+ *  +-----------------------------------------------------------------------------+
+ *  | Metric Name                  | Data Source                                  |
+ *  +------------------------------+----------------------------------------------+
+ *  |threadpool.taskCount          | {@link ThreadPoolExecutor#getTaskCount()}            |
```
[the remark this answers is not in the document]
zimmermatt: They do.

spectator-api/src/main/java/com/netflix/spectator/api/patterns/ThreadPoolMonitor.java
```diff
@@ -0,0 +1,133 @@
...
+
+  /**
+   * The default ID value.
+   */
+  private static final String DEFAULT_ID = "default";
+
+  /**
+   * Task count meter name.
+   */
+  public static final String TASK_COUNT = "threadpool.taskCount";
```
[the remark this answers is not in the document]
zimmermatt: Good point. Easier to make them public later if needed than the reverse.

spectator-api/src/main/java/com/netflix/spectator/api/patterns/ThreadPoolMonitor.java
```diff
@@ -0,0 +1,133 @@
...
+   * @param threadPool the thread pool to monitor
+   * @param threadPoolName a name with which to tag the metrics or {@code null} for the default name
+   */
+  public static void monitor(
+      final Registry registry,
+      final ThreadPoolExecutor threadPool,
+      final String threadPoolName) {
+
+    Preconditions.checkNotNull(registry, "registry");
+    Preconditions.checkNotNull(threadPool, "threadPool");
```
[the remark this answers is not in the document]
zimmermatt: Noted

spectator-api/src/test/java/com/netflix/spectator/api/patterns/ThreadPoolMonitorTest.java
zimmermatt: Reworked to use CountDownLatch.

spectator-api/src/test/java/com/netflix/spectator/api/patterns/ThreadPoolMonitorTest.java
zimmermatt: Reworked to use CountDownLatch.

zimmermatt:
So, the irony :-)

This builds and the tests run fine locally, but are failing in Travis with the CountDownLatch approach.  I'll take a look later; probably on Monday.

[Test results - Class com.netflix.spectator.api.patterns.ThreadPoolMonitorTest.pdf](https://github.com/Netflix/spectator/files/1254134/Test.results.-.Class.com.netflix.spectator.api.patterns.ThreadPoolMonitorTest.pdf)

zimmermatt:
As I've been thinking about this, I'm realizing that it's likely that the only way to make some of these tests truly deterministic (and without `Thread.sleep()`) would be to somehow introduce additional synchronization points (latches or barriers).  These sync points will need to be where the `ThreadPoolExecutor` updates the properties being monitored.

`queueSizeUpdates` and `completedTaskCountUpdates` failed across all three builds since the latched approach was introduced (894, 895, 896).

`currentThreadsBusyCountUpdates` additionally failed in the last two builds (895, 896).

brharrington:
> These sync points will need to be where the ThreadPoolExecutor updates the properties being monitored.

I think it is probably simpler to just override the methods for getting stats so they are guaranteed to be changed before the existing completed latch is changed. See sample changes pushed to branch.

zimmermatt:
Agreed, simpler than I proposed.  I wasn't even sure it was going to be possible to introduce those sync points.  The other approach I was considering was to test using a mock.  For example:

```
ThreadPoolExecutor threadPoolExecutor = Mockito.mock(ThreadPoolExecutor.class);
when(threadPoolExecutor.getActiveCount()).thenReturn(2);
etc...
```

The obvious difference is the approach you took is more of a true integration test, whereas the mock approach is more of a true unit test.

PR #464, opened 2017-09-01
Add documentation of `ThreadPoolMonitor` for the wiki.

zimmermatt:
Added some basic doc of the new API and reported metrics.

### DIFF
--- a/docs/ext/thread-pools.md
+++ b/docs/ext/thread-pools.md
@@ -1,11 +1,11 @@
 # Thread Pools
 
-Java's [ThreadPoolExecutor][ThreadPoolExecutorJavadoc] exposes several properties that are useful to monitor to assess
+Java's [ThreadPoolExecutor] exposes several properties that are useful to monitor to assess
 the health, performance, and configuration of the pool.
 
 ## Getting Started
 
-To report thread pool metrics, one can attach a [ThreadPoolMonitor][ThreadPoolMonitorJavadoc] in the following manner:
+To report thread pool metrics, one can attach a [ThreadPoolMonitor] in the following manner:
 
 ```java
 import com.netflix.spectator.api.patterns.ThreadPoolMonitor;
@@ -69,5 +69,5 @@ Gauge showing the current number of threads queued for execution.
 **Unit:** count<br/>
 **Data Source:** `ThreadPoolExecutor#getQueue().size()`
 
-[ThreadPoolExecutorJavadoc]: http://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadPoolExecutor.html
-[ThreadPoolMonitorJavadoc]: http://netflix.github.io/spectator/en/latest/javadoc/spectator-api/com/netflix/spectator/api/patterns/ThreadPoolMonitor.html
+[ThreadPoolExecutor]: http://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadPoolExecutor.html
+[ThreadPoolMonitor]: http://netflix.github.io/spectator/en/latest/javadoc/spectator-api/com/netflix/spectator/api/patterns/ThreadPoolMonitor.html
--- a/docs/ext/thread-pools.md
+++ b/docs/ext/thread-pools.md
@@ -1,0 +1,65 @@
+# Thread Pools
+
+Java's [ThreadPoolExecutor](http://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadPoolExecutor.html)
+exposes several properties that are useful to monitor to assess the health, performance, and configuration of the pool.
+
+## Getting Started
+
+To report thread pool metrics, one can attach a
+[ThreadPoolMonitor](http://netflix.github.io/spectator/en/latest/javadoc/spectator-api/com/netflix/spectator/api/patterns/ThreadPoolMonitor.html)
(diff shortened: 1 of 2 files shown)


docs/ext/thread-pools.md
```diff
@@ -0,0 +1,65 @@
+# Thread Pools
+
+Java's [ThreadPoolExecutor](http://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadPoolExecutor.html)
+exposes several properties that are useful to monitor to assess the health, performance, and configuration of the pool.
+
+## Getting Started
+
+To report thread pool metrics, one can attach a
+[ThreadPoolMonitor](http://netflix.github.io/spectator/en/latest/javadoc/spectator-api/com/netflix/spectator/api/patterns/ThreadPoolMonitor.html)
+in the following manner:
```
brharrington: Consider using reference style links to avoid having the long URIs in the middle of the paragraph when reading the raw markdown.

https://daringfireball.net/projects/markdown/syntax#link

docs/ext/thread-pools.md
```diff
@@ -0,0 +1,65 @@
...
+
+```java
+import com.netflix.spectator.api.patterns.ThreadPoolMonitor;
+import com.netflix.spectator.api.Spectator;
+
+ThreadPoolMonitor.attach(Spectator.registry(), myThreadPoolExecutor, "my-thread-pool");
+
```
brharrington: Remove empty line

docs/ext/thread-pools.md
```diff
@@ -0,0 +1,65 @@
...
+```
+
+The provided thread pool will be registered using the provided registry and polled. The metrics will be additionally
```
brharrington: > The thread pool will be polled regularly in the background and report metrics to the provided registry. The third parameter will be added to each metric as an `id` dimension. If the value is `null`, then the `default` will be used as the id.

docs/ext/thread-pools.md
```diff
@@ -0,0 +1,65 @@
...
+tagged with a key of `id` and the value of the third parameter, if provided, or a default if not.
+
+## Metrics
+
+### threadpool.taskCount
+
+Counter of the total number of tasks that have been scheduled.
+
+**Data Source:** `ThreadPoolExecutor#getTaskCount()`
```
brharrington: Please include the unit for each metric, e.g.: http://netflix.github.io/spectator/en/latest/ext/jvm-gc/#jvmgcpromotionrate

mkdocs.yml
```diff
@@ -25,6 +25,7 @@ pages:
   - 'Log4j2 Appender': ext/log4j2.md
   - 'Log4j1 Appender': ext/log4j1.md
   - 'Placeholders': ext/placeholders.md
+  - 'Thread Pools': ext/thread-pools.md
```
brharrington: For the extensions, I think we should either sort the list alphabetically or group this with other JVM stats. For example, maybe put it under "Memory Pools".

docs/ext/thread-pools.md
```diff
@@ -0,0 +1,65 @@
...
+
+To report thread pool metrics, one can attach a
+[ThreadPoolMonitor](http://netflix.github.io/spectator/en/latest/javadoc/spectator-api/com/netflix/spectator/api/patterns/ThreadPoolMonitor.html)
+in the following manner:
+
+```java
+import com.netflix.spectator.api.patterns.ThreadPoolMonitor;
+import com.netflix.spectator.api.Spectator;
+
+ThreadPoolMonitor.attach(Spectator.registry(), myThreadPoolExecutor, "my-thread-pool");
```
brharrington: `Spectator.registy()` is deprecated. Either `Spectator.globalRegistry()` or something like:

http://netflix.github.io/spectator/en/latest/ext/log4j2/#getting-started

We have generally tried to discourage use of the global registry in favor of an injected version where possible.

docs/ext/thread-pools.md
zimmermatt: Thanks; wasn't aware of that.  I've only used very simple markdown to date.

docs/ext/thread-pools.md
zimmermatt: will use `Spectator.globalRegistry()` for simplicity.  I had seen `Spectator.registy()` in the docs of JVM Buffer Pool.

docs/ext/thread-pools.md
zimmermatt: Weird... in the diff it shows an empty line, but the actual file doesn't have one.

docs/ext/thread-pools.md
brharrington: Typically you can just show something like:

```java
Registry registry = ...
ThreadPoolMonitor.attach(registry, myThreadPoolExecutor, "my-thread-pool");
```

Thanks for the pointer about buffer pool, I'll grep through the docs and get those updated.

mkdocs.yml
zimmermatt: Probably makes sense to put it with other JVM stats, but also introduce that as a parent.  Without the additional level, it makes more sense to have a flat alphabetized list.

pparth:
From what version is the thread pool monitor supported?

brharrington:
Will go out with 0.58.0.

PR #499, opened 2017-11-10
improve error message if method is not set

brharrington:
Improves the error message on HttpLogEntry to indicate
the method was not set instead of failing later when
attempting to add a null tag value to the id.

Also updates the HttpRequestBuilder to fill in the default
value for the method so it will not be null if the user
does not call `withMethod` explicitly.

/cc @copperlight


copperlight:
Awesome, thanks for the improved ergonomics!

PR #506, opened 2017-12-15
gradle 4.4 and nebula 5

brharrington:
No major improvements we care about, but keeps us in
sync with versions recommended by engtools.


brharrington:
JDK9 build is failing with:

```
org.junit.ComparisonFailure: expected:<...ion-size,1234567890,[1.0])> but was:<...ion-size,1234567890,[NaN])>
	at org.junit.Assert.assertEquals(Assert.java:115)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at com.netflix.spectator.api.CompatibilityTest.check(CompatibilityTest.java:219)
```

Not sure how the gradle version would change the behavior of that test.

brharrington:
I think this is due to GC and for some reason it is more likely to have a GC cleanup on jdk9 with the newer gradle between when the gauge is setup and when the value is polled. Note it works fine when run individually like:

```
$ ./gradlew :spectator-api:test --tests '*CompatibilityTest'
```

PR #508, opened 2018-01-12
Add gauge to Scheduler to record the timestamp of the last execution.

zimmermatt:
This commit adds monitoring of when the last execution occurred and tags
it with the result (failure if an exception was thrown, success
otherwise).  We had a case where a scheduled thread stopped running at
some point, but had no indication due to the nature of the work it does.
This will enables a "time since last success" alert.

### DIFF
--- a/spectator-api/src/main/java/com/netflix/spectator/impl/Scheduler.java
+++ b/spectator-api/src/main/java/com/netflix/spectator/impl/Scheduler.java
@@ -33,6 +33,7 @@ import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 
 /**
  * <p><b>This class is an internal implementation detail only intended for use within spectator.
@@ -96,8 +97,10 @@ public class Scheduler {
     };
   }
 
-  private static Id newId(Registry registry, String id, String name) {
-    return registry.createId("spectator.scheduler." + name, "id", id);
+  private static Id newId(Registry registry, String id, String name, String... additionalTags) {
+    return registry.createId("spectator.scheduler." + name, "id", id)
+            .withTag("id", id)
+            .withTags(additionalTags);
   }
 
   private static final Logger LOGGER = LoggerFactory.getLogger(Scheduler.class);
@@ -109,6 +112,8 @@ public class Scheduler {
   private final AtomicInteger activeCount;
   private final Timer taskExecutionTime;
   private final Timer taskExecutionDelay;
+  private final AtomicLong lastSuccessfulExecutionTimestamp;
+  private final AtomicLong lastFailedExecutionTimestamp;
   private final Counter skipped;
 
   private final ThreadFactory factory;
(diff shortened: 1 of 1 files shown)


spectator-api/src/main/java/com/netflix/spectator/impl/Scheduler.java
```diff
@@ -139,6 +144,12 @@ public Scheduler(Registry registry, String id, int poolSize) {
         .monitorValue(new AtomicInteger());
     taskExecutionTime = registry.timer(newId(registry, id, "taskExecutionTime"));
     taskExecutionDelay = registry.timer(newId(registry, id, "taskExecutionDelay"));
+    lastSuccessfulExecutionTimestamp = PolledMeter.using(registry)
+            .withId(newId(registry, id, "lastExecutionTimestamp", "result", "success"))
+            .monitorValue(new AtomicLong());
```
brharrington: This should be reporting the age not the timestamp. Also having both success/failure here doesn't really make sense to me. From the scheduler point of view it should be running regularly, the success or failure doesn't matter. Having both means normally when things are fine, if you look at the high level and do a max you'll typically get the failure since it will never get updated.

Maybe just a counter tracking uncaught exceptions that propagate. Many things will prevent the exception from throwing up so even though a task completes and doesn't throw it may still have failed from the application point of view.

spectator-api/src/main/java/com/netflix/spectator/impl/Scheduler.java
```diff
@@ -326,17 +337,28 @@ void updateNextExecutionTime(Counter skipped) {
...
      *     skipped when using {@link Policy#FIXED_RATE_SKIP_IF_LONG}.
+     * @param lastSuccessfulExecutionTimestamp
+     *     AtomicLong that will be updated to the current wallclock time of a successful run. This is useful for
+     *     monitoring.
+     * @param lastFailedExecutionTimestamp
+     *     AtomicLong that will be updated to the current wallclock time of a failed run. This is useful for
+     *     monitoring.
      */
-    void runAndReschedule(DelayQueue<DelayedTask> queue, Counter skipped) {
+    void runAndReschedule(DelayQueue<DelayedTask> queue,
```
brharrington: This should align with the margin not the opening paren. Aligning with opening braces causes future diffs to have a lot of noise just to preserve the alignment.

spectator-api/src/main/java/com/netflix/spectator/impl/Scheduler.java
zimmermatt: I'm assuming by age, you mean calculating the delta between monotonic time now and the last recorded monotonic time.

If we report the age instead of the timestamp, that won't catch the case we observed in which the task wasn't running at all... nothing would be running to update the age.  In that case it would look like the age since last success is constant and everything's okay.  I should switch to monotonic time, now that I think about it.

Maybe there's a better way to go about this.  Let's discuss in person.

Regarding the failure/exception comment... you make some good points.  I'll change it.

zimmermatt:
I was hoping to get a quick win but, after discussion, it's now clear we'll need to take a step back and assess how best to approach this.

Some initial ideas:
* report the age oldest of the tasks
* wrap it in a `LongTaskTimer`

PR #509, opened 2018-01-12
make Scheduler robust to uncaught Throwables

brharrington:
If a Throwable was propagated from an executed task,
then the thread would get stopped and never replaced.
For many uses of the scheduler, we cannot control
the task to ensure they are well-behaved and do not
want a bad task to impact other tasks that may be on
the same scheduler. With this change we catch Throwable.

An additional counter is added to track the number
of uncaught exceptions that occur for the scheduler.
This will make it easy to alert if we are seeing
regular failures.

To simplify updating the stats without having to pass
around too many handles, they have been consolidated
in a stats class.

There is also a small fix to avoid having threads
restart after shutdown. Before the shutdown was
interrupting the threads and a new one would be
created to replace them. So there would still be
threads running for the scheduler after after it
was shutdown.

### DIFF
--- a/spectator-api/src/test/java/com/netflix/spectator/impl/SchedulerTest.java
+++ b/spectator-api/src/test/java/com/netflix/spectator/impl/SchedulerTest.java
@@ -1,5 +1,5 @@
-/**
- * Copyright 2014-2016 Netflix, Inc.
+/*
+ * Copyright 2014-2018 Netflix, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -24,6 +24,7 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
 
+import java.io.IOError;
 import java.time.Duration;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ScheduledFuture;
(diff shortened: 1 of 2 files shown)


zimmermatt:
Ha! You beat me to it! Something like this was going to be my next PR. :-)

spectator-api/src/test/java/com/netflix/spectator/impl/SchedulerTest.java
```diff
@@ -88,6 +89,50 @@ public void updateNextFixedRateSkip() {
...
+
+  @Test
+  public void shutdownStopsThreads() throws Exception {
+    Scheduler s = new Scheduler(new DefaultRegistry(), "shutdown", 1);
+
+    // Schedule something to force it to start the threads
+    Scheduler.Options opts = new Scheduler.Options()
+        .withFrequency(Scheduler.Policy.FIXED_RATE_SKIP_IF_LONG, Duration.ofMillis(10))
+        .withStopOnFailure(false);
+    ScheduledFuture<?> f = s.schedule(opts, () -> {});
```
zimmermatt: Since we call `interrupt()` on each of the scheduled threads, for greater determinism consider something like:
```
    final CountDownLatch latch = new CountDownLatch(1);
    ScheduledFuture<?> f = s.schedule(opts, () -> {
      try {
        Thread.sleep(someLargeValue);
      } catch (InterruptedException e) {
        latch.countDown();        
      }
    });
```

spectator-api/src/test/java/com/netflix/spectator/impl/SchedulerTest.java
```diff
@@ -88,6 +89,50 @@ public void updateNextFixedRateSkip() {
...
+    Assert.assertEquals(1L, numberOfThreads("shutdown"));
+
+    // Shutdown and wait a bit, this gives the thread a chance to restart
+    s.shutdown();
```
zimmermatt: ... and then
```
    s.shutdown();
    Assert.assertTrue(latch.await(60, TimeUnit.SECONDS));
    Assert.assertEquals(0L, numberOfThreads("shutdown"));
```

This obviously couples the test to an implementation detail, but it seems likely calling `interrupt()` will remain.

spectator-api/src/test/java/com/netflix/spectator/impl/SchedulerTest.java
brharrington: The test reliably reproduces the problem for me, but I do like getting rid of the sleep for completing the test.

brharrington: Actually it doesn't really help. The latch is being counted down in the thread I need to check has gone away, so I still have to wait. I already know the thread has been interrupted because the shutdown method returned.

PR #512, opened 2018-01-17
fix #507, illegal reflective access on jdk9

### LINKED ISSUE #507: avoid illegal reflective access on jdk9
An optimization to reuse a character array causes warnings on jdk9:

```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.netflix.spectator.impl.AsciiSet (spectator/spectator-api/build/classes/java/main/) to constructor java.lang.String(char[],boolean)
```

Because of the new string representation, it doesn't provide the same benefit on jdk9 anyway. We should only use that method when running on jdk8.

brharrington:
The AsciiSet optimization to reuse the character array
will not only be used on jdk8. For jdk9 and later it
loses its benefit anyway because the string representation
changed to `byte[]` instead of `char[]`.

### DIFF
--- a/spectator-api/src/main/java/com/netflix/spectator/impl/AsciiSet.java
+++ b/spectator-api/src/main/java/com/netflix/spectator/impl/AsciiSet.java
@@ -17,6 +17,7 @@ package com.netflix.spectator.impl;
 
 import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodHandles;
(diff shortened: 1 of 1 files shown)


dmuino:
The code looks good but the commit message is misleading. It should say "it will only be used" instead of "it will not only be used"

brharrington:
Thanks, I'll fix it on merge.

PR #525, opened 2018-02-22
update nflx plugin to use archaius2

brharrington:
Runtime is trying to phase out the usage of archaius1.


spectator-nflx-plugin/src/main/java/com/netflix/spectator/nflx/Plugin.java
```diff
@@ -43,13 +39,11 @@
 
   /** Create a new instance. */
   @Inject
-  Plugin(Registry registry) throws IOException {
+  Plugin(Registry registry, Config config) {
```
elandau: I'd prefer to abstract out the configuration entirely so we don't force an archaius2 dependency on users.  Create an interface like this,
```
interface PluginConfig {
    boolean isEnabled();
    boolean isLoggingEnabled();
}
```

Then make archaius2 a compileOnly dependency and do some manual work to fallback from archaius's Config to properties.  So, for internal apps we will already have the binding so archaius2 will be used, but for open source apps we won't be forcing an archaius dependency on them.

```
@Provides
@Singleton
public class PluginConfigProvider implements Provider<PluginConfig> {
    @Inject
    Injector injector;
   
    @Override
    public PluginConfig get() {
        try {
            Config config = injector.getInstance(Config.class);
            return new PluginConfig() {
                @Override
                public boolean isEnabled() {
                    return config.getBoolean(ENABLED_PROP, true);
                }
                @Override
                public boolean isLoggingEnabled() {
                    return config.getBoolean(GC_LOGGING_ENABLED_PROP, true);
                }
            };
        } catch (Exception e) {
            LOGGER.warn("no archaius2 binding found, reading configuration from spectator.properties instead");
            final Properties properties = new Properties();
            try (final InputStream stream = this.getClass().getResourceAsStream("spectator.properties")) {
                properties.load(stream);
            } catch (Exception e) {
                LOGGER.warn("Unable to load spectator.properties.  Using defaults instead");
            }  
            return new PluginConfig() {
                @Override
                public boolean isEnabled() {
                    return Boolean.parseBoolean(properties.getProperty(ENABLED_PROP, "true"));
                 }
                @Override
                public boolean isLoggingEnabled() {
                     return Boolean.parseBoolean(properties.getProperty(GC_LOGGING_ENABLED_PROP, "true"));
                }
            };
        }   
    } 
}
```

Use optional bindings for PluginConfig so users can customize it to their own configuration mechanism 

`OptionalBinder.newOptionalBinder(binder(), PluginConfig.class).setDefault().toProvider(PluginConfigProvider.class)`

brharrington: > I'd prefer to abstract out the configuration entirely so we don't force an archaius2 dependency on users.

Summary of discussion with @elandau:

Longer term they are looking at also supporting Spring and those applications would not need to use archaius2 for configuration. From my point of view, the whole point of the `spectator-nflx-plugin` library is to provide the setup needed for the current standard of Guice/Governator/Archaius2. If that is not the goal, then do not install this module. For Spring specifically, we'll likely need to provide a separate setup library at some point.

For now this removes the dependency on archaius1 and we'll revisit later when we look into supporting Spring.

PR #537, opened 2018-04-17
add support for removing expired meters

brharrington:
Originally the contract for the registry was that
the meters like counter would track the state since
the meter was created. This means that meters must
stick around for the life of the process.

With some registries, e.g., AtlasRegistry, we already
removed this retriction and limited it to the count
for the last completed interval. Given we have been
running in this mode a while with no issues, we are
changing the contract to officially give the registry
implementation the choice of reset frequency. Along
with this the AbstractRegistry now has a method that
can be used to fully remove any expired meters. This
avoids a memory leak over time if the user has a bug
and the distinct meter set grows over time.

### DIFF
--- a/spectator-api/src/main/java/com/netflix/spectator/api/AbstractRegistry.java
+++ b/spectator-api/src/main/java/com/netflix/spectator/api/AbstractRegistry.java
@@ -18,10 +18,12 @@ package com.netflix.spectator.api;
 import com.netflix.spectator.api.patterns.PolledMeter;
 import com.netflix.spectator.impl.Config;
 import com.netflix.spectator.impl.Preconditions;
+import com.netflix.spectator.impl.SwapMeter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.Iterator;
+import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.function.Function;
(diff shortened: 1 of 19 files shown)


spectator-api/src/main/java/com/netflix/spectator/api/AbstractRegistry.java
```diff
@@ -221,4 +223,25 @@ private Meter compute(Meter m, Meter fallback) {
...
     return meters.values().iterator();
   }
+
+  /**
+   * Can be called by sub-classes to remove expired meters from the internal map. It will
+   * look for meters that implement {@link SwapMeter} and set the implementation to null.
+   * If user code still holds a reference to the meter, then it should retrieve a new instance
+   * from the registry on the next access.
+   */
+  protected void removeExpiredMeters() {
```
krutsko: @brharrington Could you consider to make this method public?

In our use-case, we have our own Reporter and Scheduler (rather than use built into AtlasRegistry). So, we would like AtlasRegistry can removeExpiredMeters on our request, based on our schedule.

krutsko:
Just to confirm: [Polled Gauges ](http://netflix.github.io/spectator/en/latest/intro/gauge/#polled-gauges) are never get expired, as `lastUpdated` value ticked on each poll, correct?

The only workaround I see is to use [Active Gauges](http://netflix.github.io/spectator/en/latest/intro/gauge/#active-gauges). Do you know any?

brharrington:
The lifecycle of polled gauges is tied to whatever they are polling. Spectator will keep a weak reference so if the object being polled goes away, then the polling will stop and any meters being updated by the polling will then expire.

You could also call PolledMeter.remove if the polling needs to be stopped explicitly.

PR #548, opened 2018-05-09
add max gauge type to registry

brharrington:
This was added a while back to the AtlasRegistry for
use in the aggregator service and is used internally
to timer and distribution summary implementations. To
make things more consistent it is being added to the
main registry interface.


brharrington:
Build failure is due to travis issues:

> We are now also seeing builds timing out on the sudo-enabled infrastructure (i.e. builds with `sudo: required`). We are continuing to look into it. Thank you for your patience. 
> May 9, 13:52 UTC

PR #555, opened 2018-05-22
support bounded threshold for percentile types

brharrington:
The percentile approximation types can be quite expensive
due to the number of buckets needed to get a good accuracy
for an arbitrary range. This change adds a builder for the
types that allows a threshold and accuracy to be specified.

The threshold will cap the max value for the computed data
distribution. The accuracy will be used to restrict the set
of buckets to a more reasonable number leading up to the
threshold. This is based on the common operational use case
of understanding whether the latency distribution is
approaching failure more so than actually needing an
accurate percentile. Percentiles that are closer to the
threhsold will still be accurate, but those further away
may be inflated.

/cc @kerumai


spectator-api/src/main/java/com/netflix/spectator/api/histogram/PercentileDistributionSummary.java
```diff
@@ -57,19 +73,126 @@
...
+        this.accuracy = 0.1f;
+      } else {
+        // Use the user selection.
+        this.accuracy = accuracy;
+      }
+      return this;
+    }
+
+    /**
+     * Sets the threshold for this summary. The threshold is should be the SLA boundary or
```
dmuino: `s/is should/should/`

spectator-api/src/main/java/com/netflix/spectator/api/histogram/PercentileDistributionSummary.java
```diff
@@ -57,19 +73,126 @@
...
+    public Builder withThreshold(long amount) {
+      max = amount;
+      return this;
+    }
+
+    /**
+     * Create or get an instance of the percentile distribution summary with the specified
+     * settings.
+     */
+    public PercentileDistributionSummary build() {
```
dmuino: Is it possible to refactor this to remove the duplication with `PercentileTimer#build()`?

brharrington: Probably, but it is an internal detail that we can change without impacting the API. So I'm not too worried about it right now.

spectator-api/src/main/java/com/netflix/spectator/api/histogram/PercentileDistributionSummary.java
```diff
@@ -57,19 +73,126 @@
...
+     * @param accuracy
+     *     Value from 0.0 (least accurate) to 1.0 (most accurate). Default is 0.1.
+     * @return
+     *     This builder instance to allow chaining of operations.
+     */
+    public Builder withAccuracy(float accuracy) {
+      if (accuracy < 0.0f || accuracy > 1.0f) {
+        // Invalid value provided, use default.
+        IllegalArgumentException e = new IllegalArgumentException(
+            "Invalid accuracy value for PercentileTimer [" + baseId + "]. Expected value"
```
dmuino: s/PercentileTimer/PercentileDistributionSummary/

spectator-api/src/test/java/com/netflix/spectator/api/histogram/PercentileDistributionSummaryTest.java
```diff
@@ -25,19 +27,98 @@
...
+  }
+
+  @Test
+  public void builderWithThreshold() {
+    Registry r = newRegistry();
+    PercentileDistributionSummary t = PercentileDistributionSummary.builder(r)
+        .withName("test")
+        .withThreshold(100_000)
+        .build();
+    checkPercentiles(t, 25);
```
dmuino: Is this needed because the default accuracy will make it unreliable for <25?

brharrington: Yes, those will have a higher error bound because they are less than the min value that gets selected.

brharrington:
> So, would it be possible to think about a solution that allows the developer to configure a tag name and a map of tag values and SLOs, as configuration input?

I'm not sure I follow the proposal. Can you sketch out what you would want the api to look like?

pparth:
E.g. Let's say we have a PercentileTimer for the metric requestLatency. Let's also say we have 2 resource endpoints for which we measure latency: resourceEndpointA and resourceEndpointB. Now, in your current implementation, a single SLO, a single threshold may be defined. But let's say that for resourceEndpointA a proper SLO is 10ms and for resourceEndpointB a proper SLO is 500ms. (these may be resources that belong to different services; still, we would like to have a single metric and the service name as a Tag). So, these SLOs are completely different and using a single SLO value when creating this metric, is not going to be efficient. Now, the resourceEndpoint is obviously another Tag, that, for requests routed to the first endpoint has a value of resourceEndpoint = resourceEndpointA and for requests routed to the second endpoint has a value of resourceEndpoint = resourceEndpointB. Can we somehow configure the metric to maintain different sets of buckets, based on different thresholds, per resourceEndpoint value? So, somehow configure that for the metric requestLatency, whenever we have a data point that has the resourceEndpoint Tag, and this has a value of resourceEndpointA, then a 10ms threshold and relevant bucket set should be maintained. The same with the other one. So, all we need is to define a Tag name, its presence of which, the differentiation starts, and a map of key-value pairs that define all the possible values of resourceEndpoint Tag, with the corresponding threshold.

brharrington:
You could have separate SLOs per endpoint doing something like:

```java
PercentileTimer.builder(registry)
    .withName("server.requestLatency")
    .withTag("endpoint", endpoint)
    .withThreshold(SLOs.get(endpoint), TimeUnit.MILLISECONDS)
    .build()
    .record(latency, TimeUnit.MILLISECONDS);
```

You would have to keep track of the mapping external to the timer, but I'm not sure that is a bad thing. Is there something I'm missing?

pparth:
No, looks good. We already maintain SLO mapping externally, at the configuration level.

PR #558, opened 2018-05-22
add client/server error categories to group

brharrington:
Updates the IpcErrorGroup to have general categories for
client and server errors.

### DIFF
--- a/spectator-ext-ipc/src/main/java/com/netflix/spectator/ipc/IpcErrorGroup.java
+++ b/spectator-ext-ipc/src/main/java/com/netflix/spectator/ipc/IpcErrorGroup.java
@@ -55,7 +55,17 @@ public enum IpcErrorGroup implements Tag {
   /**
    * The client is being throttled due to a problem on the server side.
    */
-  server_throttled;
+  server_throttled,
+
+  /**
+   * There was an error due to problems with the client request, e.g., HTTP status 400.
+   */
+  client_error,
+
+  /**
+   * There was an error due to problems on the server side, e.g., HTTP status 500.
+   */
+  server_error;
 
   @Override public String key() {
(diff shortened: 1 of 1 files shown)


brharrington:
I was thinking local client side internal errors fell under `general`. There is also a symmetry with the existing `client_throttling` and `server_throttling` which are both based on server responses.

kerumai:
Yes you're right about `general`. 

For the symmetry between `client_throttling` and `server_throttling` though, `client_throttling` is based on whether the local client throttled the request _before_ sending to the server, while `server_throttling` is if the server responds with a 503 status.

But on balance, I think you're correct that its worth adding `client_error`. 

So just to confirm - an example would be if server responds with a 400 status code:

ipc.client.call, ipc.error.group=client_error
ipc.server.call, ipc.error.group=client_error

And if a server responded with a 500 status code:

ipc.client.call, ipc.error.group=server_error
ipc.server.call, ipc.error.group=server_error

brharrington:
> For the symmetry between client_throttling and server_throttling though, client_throttling is based on whether the local client throttled the request before sending to the server, while server_throttling is if the server responds with a 503 status.

Oh, I interpreted `client_throttling` as being a 429 status code from the server. That is the server is saying this client needs to back off. Where a `server_throttling` or 503 says there is a problem with the server being unavailable so back off and retry on another machine.

> So just to confirm - an example would be if server responds with a 400 status code:

That looks correct to me.

kerumai:
I see - I had just modeled `client_throttling` off of `NIWSClientException.ErrorType.CLIENT_THROTTLED` which is assigned in niws if the max number of inflight requests has been reached on the _clientside_.

Anyway, I'm happy for you to merge this in whenever you're ready.

PR #559, opened 2018-05-29
use AtlasRegistry if atlas-client is new enough

brharrington:
To enable a smooth transition from ServoRegistry to
AtlasRegistry for internal users, this change detects
whether a new class in the internal client is present
and sets the registry accordingly. This should allow
us to release the newer version of Spectator and have
some users pick it up before the new version of the
client is released.


spectator-nflx-plugin/src/main/java/com/netflix/spectator/nflx/Versions.java
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * Checks that the versions for servo and atlas-client are new enough to support using
+ * the AtlasRegistry and avoiding servo. Most apps should pick up the new versions via
+ * recommendations, but there is likely to be some mismatch for a bit. This should allow
+ * a more graceful transition as it will fallback to the previous behavior if the other
+ * associated libraries are not new enough.
+ */
+class Versions {
+
+  private static Logger LOGGER = LoggerFactory.getLogger(Versions.class);
```
dmuino: missing final?

PR #580, opened 2018-06-27
Fix: don't leave totalTime at zero when using a metrics3 registry

tcataldo:
The totalTime is always zero when the registry is backed by metrics3.


brharrington:
Thanks @tcataldo ! Should go into 0.71.0 which I expect will be later this week or early next week.

PR #586, opened 2018-07-17
IpcLogger to help with using IPC metrics

brharrington:
Initial version of some basic helpers to aid in using the
common IPC metrics in a consistent manner. As an example
of using the a simple HTTP client based on the java
HttpURLConnection is instrumented. This will replace
the version from the sandbox lib.

### DIFF
--- a/spectator-ext-ipc/src/main/java/com/netflix/spectator/ipc/IpcLogEntry.java
+++ b/spectator-ext-ipc/src/main/java/com/netflix/spectator/ipc/IpcLogEntry.java
@@ -1,0 +1,1052 @@
+/*
+ * Copyright 2014-2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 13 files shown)


spectator-ext-ipc/src/main/java/com/netflix/spectator/ipc/IpcLogEntry.java
```diff
@@ -0,0 +1,1052 @@
...
+          case '\n':
+            builder.append("\\n");
+            break;
+          case '\r':
+            builder.append("\\r");
+            break;
+          case '\t':
+            builder.append("\\t");
+            break;
+          case ' ':
```
dmuino: Why is this needed?

brharrington: It shouldn't be, I think that was a hold over from a previous iteration where the whitespace would match the check in the default block.

PR #603, opened 2018-08-24
Expose sequence on ServerGroup

DanielThomas:
Turns out avoiding Frigga's `Names` class will still be useful, so thought I'd just expose the sequence here instead.


DanielThomas:
Great. Incidentally, the comment on `substr` is out of date - you're using `substring` there now.

brharrington:
Good catch on `substr`, that was a copy and paste issue. I tried a version that used CharSequence (see [here](https://github.com/Netflix/spectator/blob/master/spectator-ext-ipc/src/jmh/java/com/netflix/spectator/ipc/SeqServerGroup.java#L60)) and it was a bit faster as long as you didn't have to convert back to a string. However in all of my use-cases I pretty much had to immediately convert to a String so the benefit goes away and it was more cumbersome to use.

PR #607, opened 2018-08-31
feat(web-spring): Add PercentileTimer

ewiseblatt:
@brharrington

### DIFF
--- a/spectator-web-spring/src/main/java/com/netflix/spectator/controllers/MetricsController.java
+++ b/spectator-web-spring/src/main/java/com/netflix/spectator/controllers/MetricsController.java
@@ -16,11 +16,14 @@
 
 package com.netflix.spectator.controllers;
 
+import com.netflix.spectator.api.Counter;
+import com.netflix.spectator.api.DistributionSummary;
 import com.netflix.spectator.api.Gauge;
 import com.netflix.spectator.api.Id;
 import com.netflix.spectator.api.Measurement;
 import com.netflix.spectator.api.Meter;
 import com.netflix.spectator.api.Registry;
+import com.netflix.spectator.api.Timer;
 
 import com.netflix.spectator.controllers.filter.PrototypeMeasurementFilter;
 import com.netflix.spectator.controllers.filter.TagMeasurementFilter;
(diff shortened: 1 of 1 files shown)


ewiseblatt:
Is there any easy way I can identify a PercentileTimer?
For example in Spinnaker I am seeing things like this:

A DefaultCounter named "redis.command.latency.del"
with among other tags "statistic" with the values {"count", "totalTime", "percentile"}
and when statistic="percentile" then there is a "percentile" tag with values like T004F.

I assume this is supposed to be a PercentileTimer, but PercentileTimer is a final class, not an interface so I'm not sure what's going on. Maybe this isnt a literal PercentileTimer, but I think it is behaving like one so would like to identify the series as such.

brharrington:
PercentileTimer is an extension pattern, the registry doesn't know anything about it specifically. The pattern creates a basic timer and a set of counters that are used to model the data distribution and then compute percentile estimates at query time. I'll take a closer look at this tomorrow.

ewiseblatt:
I dont necessarily need anything changed. I'm just looking to explain why I dont get what I expected to. On my consumption end I look for and interpret the statistic tag pattern to identify Counters that are really PercentileTimers. It seems the meter itself would know this so could do so efficiently. Otherwise I dont want to add unnecessary overhead in the runtime since the consumption is essentially offline so efficiency isnt important (but would like to simplify and reduce complexity).

Also, can I rely on the interfaces to determine the types? Or maybe even the classes themselves since they are final (assuming the default registry)? It used to be that I was seeing some internal implementation classes in practice (using the default registry) but that no longer appears to be the case. I dont know if that's something I can rely on or not so I'm keeping the extra code around for future safety. If it is unnecessary, I'll cut it while I'm here.

brharrington:
> I'm just looking to explain why I dont get what I expected to. On my consumption end I look for and interpret the statistic tag pattern to identify Counters that are really PercentileTimers. It seems the meter itself would know this so could do so efficiently.

PercentileTimer/PercentileDistributionSummary are not core types in the registry. They are patterns built on top of the registry that generate a set of core meters. In the case of PercentileTimer, it will create a Timer and a set of Counters that are used to model the data distribution. So you will not see the PercentileTimer type when querying the registry.

> Also, can I rely on the interfaces to determine the types?

For core types, yes you should be able to rely on the interface. Specific implementations of those cannot be relied on as those can change depending on the registry being used.

PR #619, opened 2018-09-13
enable CI build for jdk11

brharrington:
Now that there is an updated spotbugs release we should
be able to build on jdk11. Dropping jdk10 as it will no
longer be supported soon.


brharrington:
https://jdk.java.net/ is down, need to wait for it to be up so that we can resolve the EA openjdk11 build.

PR #624, opened 2018-10-03
Ensure the AtlasRegistry uses a proper config

dmuino:
Replicate the behavior used internally at Netflix by attempting to load
account and environment specific properties, and populating common-tags
so subscriptions can be properly evaluated.

### DIFF
--- a/spectator-nflx-plugin/src/main/java/com/netflix/spectator/nflx/NetflixConfig.java
+++ b/spectator-nflx-plugin/src/main/java/com/netflix/spectator/nflx/NetflixConfig.java
@@ -1,0 +1,209 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


spectator-nflx-plugin/src/main/java/com/netflix/spectator/nflx/SpectatorModule.java
```diff
@@ -110,15 +111,20 @@
...
     @Inject(optional = true)
     private Clock clock;
 
     Config config() {
-      if (config == null) {
-        LOGGER.warn("no archaius2 binding found, using empty configuration");
-        config = EmptyConfig.INSTANCE;
+      if (atlasConfig == null) {
+        if (config == null) {
+          LOGGER.warn("no archaius2 binding found");
```
brharrington: It looks like this will already get logged as part of `createConfig`. Do we need both?

spectator-nflx-plugin/src/main/java/com/netflix/spectator/nflx/NetflixConfig.java
```diff
@@ -0,0 +1,209 @@
...
+import java.net.UnknownHostException;
+import java.util.Enumeration;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Properties;
+
+/**
+ * Utility class for creating a configuration to be used in the Netflix environment.
+ */
+public final class NetflixConfig {
```
brharrington: I guess this needs to be public to reuse with some internal use-cases, but we should update the comment to note it is considered an internal detail and should not be relied on by others.

PR #631, opened 2018-10-24
update IPC metrics to use status and detail

brharrington:
Updates the IPC metrics pattern based on recent internal
conversations. In particular, many found the errorGroup
and errorReason values to be confusing and these have been
replaced by status and statusDetail.


spectator-ext-ipc/src/main/java/com/netflix/spectator/ipc/IpcTagKey.java
```diff
@@ -144,7 +166,11 @@
 
   /**
    * HTTP status code.
+   *
+   * @deprecated Use {@link #statusDetail} instead. This value will be removed in
```
kerumai: What was the reason for this? I'd rather keep this than start putting it into statusDetail.

PR #640, opened 2018-11-01
registry that delegates to Micrometer

brharrington:
Wraps the Micrometer MeterRegistry interface so that code
instrumented using the Spectator API can use Micrometer
registry implementations. Micrometer API borrowed a lot
from Spectator so most of the concepts translate directly.

/cc @ewiseblatt @jkschneider


jkschneider:
Thanks for tackling this @brharrington. LGTM.

PR #654, opened 2018-12-12
provide access to log entry from request builder

brharrington:
For some use-cases it can be useful to perform some minor
adjustments to the log entry for the request.

/cc @rnz0


rnz0:
LGTM

PR #670, opened 2019-01-19
Allow specifying custom tags for AWS metrics collector

akhaku:


spectator-ext-aws/src/main/java/com/netflix/spectator/aws/SpectatorRequestMetricCollector.java
```diff
@@ -72,37 +72,47 @@
...
+  }
+
+  /**
+   * Constructs a new instance. Custom tags provided by the user will be applied to every metric.
+   * If there is a conflict between a tag added by SpectatorRequestMetricCollector and a custom tag,
+   * the custom tag will override the built-in one.
+   */
+  public SpectatorRequestMetricCollector(Registry registry, Map<String, String> customTags) {
     this.registry = Preconditions.checkNotNull(registry, "registry");
+    this.customTags = Preconditions.checkNotNull(customTags, "customTags");
```
akhaku: A couple other options:
- provide a setter, so that a user can change the tags sent after construction
- a protected method that can be overridden by subclassing (eg making getAllTags protected to it can be overridden), which would allow users to specify custom logic to tag their metrics with
The latter option is most powerful - thoughts?

brharrington: The map would need to get copied to ensure it cannot be mutated after being passed in.

Either option you list there would have additional risks of thread safety issues.

spectator-ext-aws/src/main/java/com/netflix/spectator/aws/SpectatorRequestMetricCollector.java
```diff
@@ -72,37 +72,48 @@
...
    * Constructs a new instance.
    */
   public SpectatorRequestMetricCollector(Registry registry) {
+    this(registry, Collections.emptyMap());
+  }
+
+  /**
+   * Constructs a new instance. Custom tags provided by the user will be applied to every metric.
+   * If there is a conflict between a tag added by SpectatorRequestMetricCollector and a custom tag,
+   * the custom tag will override the built-in one.
```
brharrington: Why is this desirable? Typically the built-in values are extracted from the context of the request, since you cannot know this for a static map, it seems strange to allow that sort of override. Perhaps a better option would be a static check that would reject custom tags that try to use a key that is provided by default.

akhaku:
Generally speaking, the use case is for tagging these metrics to make them easier to find across a group of apps, where some apps may be using our library to talk to S3 and some are using their own. As a library owner (where the library uses the S3 SDK), I'm looking for a way to search for these metrics across all apps, and to be sure that they're using my library so I can scope my search to those apps.

spectator-ext-aws/src/main/java/com/netflix/spectator/aws/SpectatorRequestMetricCollector.java
akhaku: Ah right, thanks.

spectator-ext-aws/src/main/java/com/netflix/spectator/aws/SpectatorRequestMetricCollector.java
akhaku: I don't see a current use case for it, but the idea was to provide some sort of extensibility, where a user could choose to override an existing tag to eg repurpose it for something else.
If that's not something we want, the static check makes sense.

akhaku: You know, if we added a static check, it makes it more difficult to add tags in the future, because we could break people at runtime. That doesn't seem like a good idea. I think we should pick a default - have the user's tags override ours or vice versa. My thoughts on this are that if a user is trying to override one of our tags then we should let them, so I think the former makes sense. What do you think?

brharrington: Permitting customization at all means we potentially break users if the default set of tags from the extension changes. If allowing customization, it is more of a question of how things break. In many cases where there is a concern for the fall out of runtime errors the behavior is to log by default, but throw for test cases:

http://netflix.github.io/spectator/en/latest/intro/testing/#exceptions

The goal is to make it more likely the developer will notice early on and at least log so that it is not completely silent when something is misbehaving.

Regarding overrides, in my opinion a user shouldn't override any of the provided tags because it causes confusion and makes it harder to reason about aggregate behavior using these metrics. For example, if the user overrides the `requestType`, then everyone querying it would need to know that or they could be confused that they are not seeing the contribution for a subset of the requests because the developer chose to put some arbitrary value there.

brharrington:
Thanks for providing the use-case. Note, you would have to explicitly specify the request collector for the client used by your library on on each request. The most common usage is for it to just configure it using `AwsSdkMetrics.setMetricCollector`, but that will not work for the scoping you are trying to accomplish. I did confirm that setting a collector at the client or request level will take precedence and will not result in double counting.

spectator-ext-aws/src/main/java/com/netflix/spectator/aws/SpectatorRequestMetricCollector.java
akhaku: Synced up in person, makes sense, agreed. Thanks for pointing me at Registry#propagate 👍

PR #681, opened 2019-02-14
add reset() method for DefaultRegistry

brharrington:
The use-case is to allow for resetting the state between
test invocations when the registry instance needs to be
reused.

/cc @quidryan


quidryan:
Thanks!

PR #682, opened 2019-02-15
Doc for the Common IPC Metrics extension

kerumai:


docs/ext/ipc.md
```diff
@@ -0,0 +1,200 @@
...
+  the server (eg. the URL path prefix for a java servlet, or the grpc endpoint name).
+* `ipc.attempt`: Which attempt at sending this message is this. Allowed Values =
+  [`initial`, `second`, `third_up`] (`initial` is the first attempt, `second` is 2nd attempt
+  but first *retry*, `third_up` means third or higher attempt).
+* `ipc.attempt.final`: Indicates if this request was the final attempt of potentially multiple
+  retry attempts. Allowed Values = [`true`, `false`].
+* `ipc.server.app`: The `nf.app` of the server the message is being sent *to*.
+* `ipc.server.asg`: The `nf.asg` of the server the message is being sent *to*.
+* `ipc.client.app`: The `nf.app` of the server the message is being sent *from*.
+* `ipc.client.asg`: The `nf.asg` of the server the message is being sent *from*.
```
brharrington: Were `ipc.server.cluster` and `ipc.client.cluster` omitted on purpose?

docs/ext/ipc.md
```diff
@@ -0,0 +1,200 @@
...
+IPC messages from a client.
+
+**Dimensions:**
+
+* `ipc.protocol`
+* `ipc.vip`
+* `ipc.endpoint`
+* `ipc.result`
+* `ipc.status`
+* `ipc.status.detail`
```
brharrington: Status/result should be remove here correct? Since the inflight metric needs to get updated before the result is received, I don't think it is possible to include that information on the inflight tracker.

kerumai: Yes, my mistake.

docs/ext/ipc.md
kerumai: They were, although maybe they shouldn't be. It's because I omitted them in the original proposal doc (with the thinking being that the minimum of `nf.asg` was needed), but then I ended up adding cluster in the implementations I worked on ...  so maybe the spec should follow the implementations here and we should document these?

brharrington: I think they should be included in the docs if they will show up on the metrics.

PR #692, opened 2019-03-04
Spring Boot 2 properties compatibility

robzienert:
Updates the MetricsController spring properties to be compatible with Spring Boot 2. Boot 2 enforces convergence on kebab-case configurations within code. User-land configuration will continue to resolve correctly if kept as camelCase.

This is to support Spinnaker's upgrade to Spring Boot 2.


brharrington:
@robzienert This was released in 0.87.0.

PR #736, opened 2019-07-17
avoid exceptions if map is modified concurrently

### LINKED ISSUE #733: ArrayTagSet#addAll(Map<String, String> ts) can throw ArrayOutOfBoundsException
Using version 0.92.1 I ran into an ArrayIndexOutOfBoundsException while adding tags stored in a ConcurrentMap.  This can happen if an item is added after the array is created but before the map is iterated on.  

```
java.lang.ArrayIndexOutOfBoundsException: 1
        at com.netflix.spectator.api.ArrayTagSet.addAll(ArrayTagSet.java:169)
        at com.netflix.spectator.api.DefaultId.withTags(DefaultId.java:71)
        at com.netflix.spectator.api.DefaultId.withTags(DefaultId.java:26)
 ```

brharrington:
This change special cases `ConcurrentMap` to ensure that the
code will not throw if there is a concurrent modification
while it is being added to the list. A concurrent modification
should be considered a bug in the caller, but it should not
propagate and cause a bigger problem for the app. Also it
is not possible to detect all cases, only ones that fail in
a noticeable way.

The docs for `Id` have also been updated to clarify the
expectation for the user.

Fixes #733.

/cc @elandau

### DIFF
--- a/spectator-api/src/main/java/com/netflix/spectator/api/ArrayTagSet.java
+++ b/spectator-api/src/main/java/com/netflix/spectator/api/ArrayTagSet.java
@@ -21,6 +21,7 @@ import java.util.Collection;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentMap;
 
 /**
  * An immutable set of tags sorted by the tag key.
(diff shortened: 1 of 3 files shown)


elandau:
Thanks for the quick turnaround 👍

PR #737, opened 2019-07-29
Make Registry and Meter Serializable

joan38:
Hi @brharrington, in DSE we are trying to run this thing in Flink.
Any way to make those Serializable?


joan38:
Actually this is not going to work as we would need to make a lot more Serializable

brharrington:
File an issue if you want us to look into it. I know some users have used it with Spark which I would think would have the same problem, but I would have to better understand the details of how they were using it.

PR #761, opened 2019-09-23
update travis build to include jdk13

brharrington:
Drop jdk12 and add jdk13.


brharrington:
Currently blocked by https://github.com/gradle/gradle/issues/8681.

PR #768, opened 2019-10-15
Update IpcTagKey.java

haleyw:
Add the missing failure injected tag along with documentation.

### DIFF
--- a/spectator-ext-ipc/src/main/java/com/netflix/spectator/ipc/IpcTagKey.java
+++ b/spectator-ext-ipc/src/main/java/com/netflix/spectator/ipc/IpcTagKey.java
@@ -163,6 +163,13 @@ public enum IpcTagKey {
    * The port number connected to on the server.
    */
   serverPort("ipc.server.port"),
+  
+  /**
+   * Indicates that an artificial failure was injected into the request processing for testing purposes. 
+   * The outcome of that failure will be reflected in the other error tags. 
+   * Valid values are {@code none}, {@code failure}, {@code delay}.
+   */
+  failureInjected("ipc.failure.injected"),
 
   /**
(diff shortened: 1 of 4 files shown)


spectator-ext-ipc/src/main/java/com/netflix/spectator/ipc/IpcTagKey.java
```diff
@@ -163,6 +163,13 @@
    * The port number connected to on the server.
    */
   serverPort("ipc.server.port"),
+  
+  /**
+   * Indicates that an artificial failure was injected into the request processing for testing purposes. 
+   * The outcome of that failure will be reflected in the other error tags. 
+   * Valid values are {@code none}, {@code failure}, {@code delay}.
```
brharrington: If the set of values is fixed, should there be a dedicated enum for that type that can be used? Similar to IpcResult or IpcStatus:

https://github.com/Netflix/spectator/blob/master/spectator-ext-ipc/src/main/java/com/netflix/spectator/ipc/IpcResult.java

This would also allow us to integrate checks for this into the validation to help better ensure consistency going forward:

https://github.com/Netflix/spectator/blob/master/spectator-ext-ipc/src/main/java/com/netflix/spectator/ipc/IpcMetric.java#L182

haleyw: Ahh, yeah, that could be useful. Will make the updates -- thanks!

haleyw: I moved the valid values into an enum, enabled validation, and added a couple of tests.

PR #775, opened 2019-10-31
optimize the collection of measurements

brharrington:
Removes the allocation of temporary lists from the
measure call. Also reduces some of the checks for
each meter:

- Expiration check can be avoided because the expired
  meters will get removed in the context where the
  data is collected.
- NaN checks inlined to the gauge type which is the
  only type where it would be possible to have a NaN
  value for the measurement.

### DIFF
--- a/spectator-reg-atlas/src/main/java/com/netflix/spectator/atlas/AtlasMaxGauge.java
+++ b/spectator-reg-atlas/src/main/java/com/netflix/spectator/atlas/AtlasMaxGauge.java
@@ -22,7 +22,7 @@ import com.netflix.spectator.api.Measurement;
 import com.netflix.spectator.api.Statistic;
 import com.netflix.spectator.impl.StepDouble;
 
-import java.util.Collections;
+import java.util.List;
 
 /**
  * Gauge that reports the maximum value submitted during an interval to Atlas. Main use-case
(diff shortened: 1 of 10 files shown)


spectator-reg-atlas/src/main/java/com/netflix/spectator/atlas/AtlasMaxGauge.java
```diff
@@ -43,18 +43,19 @@
...
     double v = value.poll();
-    final Measurement m = new Measurement(stat, value.timestamp(), v);
-    return Collections.singletonList(m);
+    ms.add(new Measurement(stat, value.timestamp(), v));
   }
 
   @Override public void set(double v) {
-    value.getCurrent().max(v);
-    updateLastModTime();
+    if (Double.isFinite(v)) {
```
dmuino: Why does max gauge do the check on set and a regular gauge on measure?

brharrington: Max is activity based and automatically resets to 0.0. This is preventing a single bad recorded value from corrupting the max. Though double checking, this can be removed because the max function already has a sanity check. It is similar to the checks to exclude negative values on timers.

For a regular gauge, it will just report the last value until the TTL. Setting explicitly to `NaN` is common for clearing without waiting so in the measure we just double check and do not emit if it would get dropped later anyway.

PR #784, opened 2020-01-13
customizable retry policy for http client

brharrington:
Updates the sample HTTP client that wraps HttpURLConnection
to support a RetryPolicy. By default it uses a safe policy
that will retry read requests for all exceptions, throttling,
and 5xx errors. Write requests will only get retried for
connect exceptions and throttling. Other failures could have
a partial write and are considered unsafe to retry.


dmuino:
Looks good to me. The only minor issue I see is that it's not possible with the current implementation to have a `RetryPolicy` that will not retry 503s for example, since 429/503s are hard-coded to retry with a delay. I don't see this as a problem though but it might be worth documenting the behavior.

brharrington:
Thanks, I updated so the policy controls whether it will retry for throttling as well.

PR #791, opened 2020-01-15
add delay_and_failure failure injection type

snowp:
This allows properly tagging IPC metrics where both a failure
and latency has been injected.


brharrington:
I restarted the failed build, I think that is an unrelated test that is flakey.

PR #803, opened 2020-03-18
add jdk14 to build matrix

brharrington:
Also drop jdk13 as we only commit to supporting 2 LTS and
latest.


brharrington:
Will need to wait for gradle 6.3: https://github.com/gradle/gradle/issues/10248

PR #827, opened 2020-06-29
batch payloads to LWC API service

brharrington:
Batch the data going to LWC API. This change reuses the
batch size parameter for payloads going to the publish
endpoint. Ensures that the max request size is bounded.

### DIFF
--- a/spectator-reg-atlas/src/main/java/com/netflix/spectator/atlas/impl/EvalPayload.java
+++ b/spectator-reg-atlas/src/main/java/com/netflix/spectator/atlas/impl/EvalPayload.java
@@ -15,6 +15,7 @@
  */
 package com.netflix.spectator.atlas.impl;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


spectator-reg-atlas/src/main/java/com/netflix/spectator/atlas/impl/EvalPayload.java
```diff
@@ -59,6 +60,31 @@ public long getTimestamp() {
...
+   *     Size of the metric batches to create.
+   * @return
+   *     List of payloads that have at most {@code batchSize} metrics per payload.
+   */
+  public List<EvalPayload> toBatches(int batchSize) {
+    int size = metrics.size();
+    if (size <= batchSize) {
+      return Collections.singletonList(this);
+    } else {
+      List<EvalPayload> payloads = new ArrayList<>(size / batchSize + 2);
```
jfz, line 77: Not +1 here?

brharrington: Good catch, the +2 was from an earlier attempt where I was looking at splitting the messages as well. Decided it wasn't worth the complexity right now with current usage.

PR #869, opened 2021-03-03
Serializable Registry

joan38:


brharrington:
Why do you want the registry to be serializable?

joan38:
Actually I found a PR I've done in the past https://github.com/Netflix/spectator/pull/737
We are using this in Flink and things need to be serializable. We have a workaround but I was wondering if we can remove it.

In that PR I said that it's not going to work. I haven't dived into this further but I'm wondering if this requires a larger change to make it serializable.

brharrington:
Registry is just a base interface, all implementations would have to be made serializable. In general I don't think it is a good idea as things like gauge polling allow registering callbacks for interrogating the state of arbitrary objects. Also the registry will typically have context for the local instance and should normally be a singleton in the process. Running multiple instances or transferring to another instance would not work correctly.

joan38:
That make sense. I'm closing this one for now.

PR #885, opened 2021-05-13
deprecate IpcTagKey entries that are not in spec

brharrington:
Consistency should be a high priority for the common IPC
metrics and we should discourage additions that are not
part of the [spec]. As a first step, the existing entries
for the IpcTagKey enum that are not in the spec have been
deprecated.

[spec]: https://netflix.github.io/spectator/en/latest/ext/ipc/

### DIFF
--- a/spectator-ext-ipc/src/main/java/com/netflix/spectator/ipc/IpcLogEntry.java
+++ b/spectator-ext-ipc/src/main/java/com/netflix/spectator/ipc/IpcLogEntry.java
@@ -641,15 +641,11 @@ public final class IpcLogEntry {
       putTag(tags, getAttemptFinal());
 
       // Optional for client
-      putTag(tags, IpcTagKey.serverRegion.key(), serverRegion);
-      putTag(tags, IpcTagKey.serverZone.key(), serverZone);
       putTag(tags, IpcTagKey.serverApp.key(), serverApp);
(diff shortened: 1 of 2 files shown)


jleibund:
One question from [886](https://github.com/Netflix/spectator/pull/886) about httpMethod and httpStatus:

- Currently our gRPC doesn't report either of these
- Although our gRPC is H2 method is only used, in practice in the swagger bridge and always POST
- Also re gRPC the httpStatus isn't meaningful when compared with gRPC status codes
- In similar fashion GraphQL doesn't leverage http status codes except for network errors - they are tunneled through 200 OK responses
- It seems to me that http codes are valuable mostly for http-centric comms
- If IPC metrics are truly common, what of UDP (DNS) etc or websockets

Just wondering, as part of this, can we document or annotate the httpStatus and httpMethod tags as applicable to http communication - some sort of caveat that they should be required when http or http-based but optional where less meaningful or when tagging a non-http (or even non-tcp) based IPC protocol?

brharrington:
Yeah, the http status/method stuff are optional and only intended for HTTP based requests. That should be clarified in the spec.

kerumai (approved):
This seems reasonable. You may need to communicate this out to people/apps that currently publish the removed tags, before releasing this though.

PR #889, opened 2021-06-29
ipc: add helper for sanitizing paths

brharrington:
Adds a helper class to sanitize URL paths before using
as a tag value. This is best effort and tries to do a
reasonable job for common patterns that are seen. Users
would need to manually customize before calling if more
specific mappings are necessary.


dmuino (approved):
This is very good. I really like the way you're detecting random looking segments, and I might steal that for the our logging counterpart.

jleibund (approved):

PR #894, opened 2021-08-12
reduce http logging

copperlight:
Only log the last failed HTTP attempt at warn level, to reduce noise in the
logging platform.

### DIFF
--- a/spectator-ext-ipc/src/main/java/com/netflix/spectator/ipc/http/HttpRequestBuilder.java
+++ b/spectator-ext-ipc/src/main/java/com/netflix/spectator/ipc/http/HttpRequestBuilder.java
@@ -293,7 +293,7 @@ public class HttpRequestBuilder {
         if (attempt == numAttempts || !retryPolicy.shouldRetry(method, e)) {
           throw e;
         } else {
-          LOGGER.warn("attempt {} of {} failed: {} {}", attempt, numAttempts, method, uri);
+          LOGGER.info("attempt {} of {} failed: {} {}", attempt, numAttempts, method, uri);
         }
       }
     }


spectator-ext-ipc/src/main/java/com/netflix/spectator/ipc/http/HttpRequestBuilder.java
```diff
@@ -293,7 +293,11 @@ public HttpResponse send() throws IOException {
         if (attempt == numAttempts || !retryPolicy.shouldRetry(method, e)) {
           throw e;
         } else {
-          LOGGER.warn("attempt {} of {} failed: {} {}", attempt, numAttempts, method, uri);
+          if (attempt < numAttempts) {
+            LOGGER.info("attempt {} of {} failed: {} {}", attempt, numAttempts, method, uri);
+          } else {
```
lavanyachennupati, line 298: I think this `else` condition is unreachable. Or am I missing something?

copperlight: Yep, good catch. Fixed.

spectator-ext-ipc/src/main/java/com/netflix/spectator/ipc/http/HttpRequestBuilder.java
```diff
@@ -291,9 +291,10 @@ public HttpResponse send() throws IOException {
         }
       } catch (IOException e) {
         if (attempt == numAttempts || !retryPolicy.shouldRetry(method, e)) {
+          LOGGER.warn("attempt {} of {} failed: {} {}", attempt, numAttempts, method, uri);
```
brharrington, line 294: At this point the exception is bubbled up to the caller to log or otherwise handle with additional context. I don't think we should be adding what is likely a duplicate warning here.

spectator-ext-sandbox/src/main/java/com/netflix/spectator/sandbox/HttpRequestBuilder.java
```diff
@@ -231,9 +231,10 @@ public HttpResponse send() throws IOException {
         // debatable, but we have seen them in some cases if there is a high latency for
         // DNS lookups. So for now assume all exceptions are transient issues.
         if (attempt == numAttempts) {
+          LOGGER.warn("attempt {} of {} failed: {} {}", attempt, numAttempts, method, uri);
           throw e;
         } else {
-          LOGGER.warn("attempt {} of {} failed: {} {}", attempt, numAttempts, method, uri);
+          LOGGER.info("attempt {} of {} failed: {} {}", attempt, numAttempts, method, uri);
```
brharrington, line 237: The sandbox client shouldn't get used anymore and is only still around just in case for backwards compatibility, we don't need to be maintaining it or pushing new changes there. In general the sandbox sub-project is to incubate things and there is no guarantee of stability.

copperlight: Very good, will back out that part of the change.

brharrington (approved):

spectator-ext-ipc/src/main/java/com/netflix/spectator/ipc/http/HttpRequestBuilder.java
copperlight: Fair. Removed.

PR #918, opened 2021-10-28
collect connection pool metrics from aws

dbyron-sf:


spectator-ext-aws/src/main/java/com/netflix/spectator/aws/SpectatorRequestMetricCollector.java
```diff
@@ -94,10 +102,29 @@ public SpectatorRequestMetricCollector(Registry registry) {
...
+   */
+  public SpectatorRequestMetricCollector(Registry registry, HandlerContextKey<String> handlerContextKey) {
+    this(registry, Collections.emptyMap(), handlerContextKey);
+  }
+
+  /**
+   * Constructs a new instance using no handler context key and custom tags.
    */
   public SpectatorRequestMetricCollector(Registry registry, Map<String, String> customTags) {
+    this(registry, customTags, null);
```
dbyron-sf, line 116: @brharrington Actually as I stare at this it would be really helpful for all the bean wiring, etc. to have a default context key.  What do you think of e.g. https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/handlers/HandlerContextKey.html#OPERATION_NAME?

dbyron-sf: Here goes: https://github.com/Netflix/spectator/pull/918/commits/c089bbbc96f0257788e377b9b4305c204b13fc2c

brharrington: Operation name seems reasonable.

spectator-ext-aws/src/main/java/com/netflix/spectator/aws/SpectatorRequestMetricCollector.java
```diff
@@ -156,6 +203,10 @@ private Id metricId(String metric, Map<String, String> tags) {
       allTags.put(tag.getName(), tag.getValue(metrics).orElse(UNKNOWN));
     }
     allTags.put(TAG_REQUEST_TYPE, request.getOriginalRequest().getClass().getSimpleName());
+    String contextTagValue = request.getHandlerContext(handlerContextKey);
+    if (contextTagValue != null) {
```
brharrington, line 207: Should the gauges be contingent on the context being set? Otherwise existing users might start getting the gauges for pool metrics and not realize that the values reported do not make sense when they have multiple pools in use.

dbyron-sf: Yeah, that makes sense: https://github.com/Netflix/spectator/pull/918/commits/5b77464959b3aa7166c30f601b1eead476cc8750

brharrington (approved):

dbyron-sf:
Thanks much!

PR #922, opened 2021-11-10
fix(aws): change default handler context key

dbyron-sf:
The aws sdk places a value in HandlerContextKey.OPERATION_NAME (e.g. GetObject) at least
some of the time, so use a default that's not one of the predefined HandlerContextKey
objects to avoid that.

I noticed this with a test of spectator 1.0.5 against a localstack container.  Happy to add a test like that here is it's OK to add a test dependency on https://github.com/testcontainers/testcontainers-java.  Sorry I didn't catch this earlier.

### DIFF
--- a/spectator-ext-aws/README.md
+++ b/spectator-ext-aws/README.md
@@ -86,10 +86,10 @@ that caused the throttling to occur.
 
 To help distinguish metrics from multiple clients, it's possible to specify a
 [HandlerContextKey](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/handlers/HandlerContextKey.html)
-for a SpectatorRequestMetricCollector (HandlerContextKey.OPERATION_NAME by
-default).  The SpectatorRequestMetricCollector looks for a value from each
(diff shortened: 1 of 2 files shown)


brharrington (approved):

brharrington:
> Happy to add a test like that here is it's OK to add a test dependency on https://github.com/testcontainers/testcontainers-java

I'm not familiar with this, the main thing I would want to make sure is that it is fairly lightweight, reliable, and wouldn't significantly increase testing times.  It can be done as a follow up later if there is sufficient need or interest.

dbyron-sf:
> > Happy to add a test like that here is it's OK to add a test dependency on https://github.com/testcontainers/testcontainers-java
> 
> I'm not familiar with this, the main thing I would want to make sure is that it is fairly lightweight, reliable, and wouldn't significantly increase testing times. It can be done as a follow up later if there is sufficient need or interest.

Yeah, it would for sure increase testing times...probably significantly...time to pull a docker image, and then run it.

dbyron-sf:
Thanks for merging this and releasing a new version so quickly!

PR #938, opened 2022-01-31
reduce calls to access time on record

brharrington:
Update the Atlas distribution summary and timer so that
the record calls will only access the time value once for
a record call. Before it would happen internally to the
step value for each summary statistic. This can add up if
the meter is updated from a tight loop.


jasonk000 (approved):
Thanks!

PR #941, opened 2022-01-31
support batch record() of metrics in Timer and DistributionSummary

jasonk000:
Supports record of many Timer and DistributionSummary metrics in a single call, and provides an optimised implementation for the Atlas implementation.

### DIFF
--- a/spectator-api/src/main/java/com/netflix/spectator/api/ManualClock.java
+++ b/spectator-api/src/main/java/com/netflix/spectator/api/ManualClock.java
@@ -25,6 +25,7 @@ public class ManualClock implements Clock {
 
   private final AtomicLong wall;
   private final AtomicLong monotonic;
+  private final AtomicLong countPolled;
 
   /** Create a new instance. */
   public ManualClock() {
@@ -42,14 +43,21 @@ public class ManualClock implements Clock {
   public ManualClock(long wallInit, long monotonicInit) {
     wall = new AtomicLong(wallInit);
     monotonic = new AtomicLong(monotonicInit);
+    countPolled = new AtomicLong(0);
   }
 
   @Override public long wallTime() {
+    countPolled.incrementAndGet();
     return wall.get();
   }
 
   @Override public long monotonicTime() {
+    countPolled.incrementAndGet();
     return monotonic.get();
+  }
+
+  public long countPolled() {
+    return countPolled.get();
   }
 
   /** Set the wall time to the value {@code t}. */
--- a/spectator-api/src/main/java/com/netflix/spectator/api/Timer.java
+++ b/spectator-api/src/main/java/com/netflix/spectator/api/Timer.java
@@ -52,6 +52,50 @@ public interface Timer extends Meter {
   }
 
   /**
+   * Updates the statistics kept by the counter with the specified amount. Behaves as if
+   * `record()` was called in a loop, but may be faster in some cases.
+   *
+   * @param amounts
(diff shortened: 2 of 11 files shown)


spectator-api/src/main/java/com/netflix/spectator/api/Timer.java
```diff
@@ -51,6 +51,50 @@ default void record(Duration amount) {
     record(amount.toNanos(), TimeUnit.NANOSECONDS);
   }
 
+  /**
+   * Updates the statistics kept by the counter with the specified amount. Behaves as if
```
brharrington, line 55: > Updates the statistics kept by the **timer** with the specified **amounts**.

spectator-api/src/main/java/com/netflix/spectator/api/Timer.java
```diff
@@ -51,6 +51,50 @@ default void record(Duration amount) {
...
+   */
+  default void record(long[] amounts, int n, TimeUnit unit) {
+    final int limit = Math.min(amounts.length, n);
+    for (int i = 0; i < limit; i++) {
+      record(amounts[i], unit);
+    }
+  }
+
+  /**
+   * Updates the statistics kept by the counter with the specified amount. Behaves as if
```
brharrington, line 78: > Updates the statistics kept by the **timer** with the specified **amounts**.

spectator-reg-atlas/src/test/java/com/netflix/spectator/atlas/AtlasTimerTest.java
```diff
@@ -144,6 +145,116 @@ public void recordSeveralValues() {
...
+    checkValue(0, 0, 0, 0);
+
+    clock.setWallTime(step + 1);
+    checkValue(1, 2, 4, 2);
+  }
+
+
+  @Test
+  public void recordBatchOverflowSingle()
+  {
```
brharrington, line 197: formatting, please move brace to end of previous line to be consistent with other code

spectator-reg-atlas/src/test/java/com/netflix/spectator/atlas/AtlasTimerTest.java
```diff
@@ -144,6 +145,116 @@ public void recordSeveralValues() {
...
+      dist.record(new long[]{ amount }, 1, TimeUnit.NANOSECONDS);
+      square += (double) amount * amount;
+    }
+    clock.setWallTime(step + 1);
+    checkValue(10000, 10e3 * amount, square, amount);
+  }
+
+  @Test
+  public void recordBatchOverflowBatch()
+  {
```
brharrington, line 212: formatting

spectator-api/src/main/java/com/netflix/spectator/api/ManualClock.java
```diff
@@ -42,16 +43,23 @@ public ManualClock() {
...
+    countPolled.incrementAndGet();
     return wall.get();
   }
 
   @Override public long monotonicTime() {
+    countPolled.incrementAndGet();
     return monotonic.get();
   }
 
+  public long countPolled() {
```
brharrington, line 59: I would have to think about this as it would be a new public user-facing API. I'm not sure how often this will come up as we haven't used it before. Also, if it does come up again it might be necessary to separate out which timing call was made.

How about just creating a custom sub-class that tracks this and is part of the test cases for now. We can revisit in the future if there is enough need.

brharrington (approved):

PR #950, opened 2022-03-22
batch increment for BucketCounter

brharrington:
Add `increment` method to BucketCounter to allow for
batch increments of the same amount.


manolama (approved):
Do we need to update the other languages as well?

PR #954, opened 2022-04-04
Additional JVM metrics

DanielThomas:
I was looking through our current JVM metrics and identified some areas where we have gaps:

- ClassLoading - Classloading is the most significant contributor to startup time and we've seen apps affected by runtime class generation issues more than once
- Thread - Characterize JVMs by thread count, detect issues with excessive thread allocations
- Compilation - Understand compilation times particularly when evaluating alternative JVMs

I'll start a thread internally to discuss.

### DIFF
--- a/spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
+++ b/spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
@@ -44,7 +44,7 @@ public final class Jmx {
     monitorThreadMXBean(registry);
     monitorCompilationMXBean(registry);
 
-    for (MemoryPoolMXBean mbean : ManagementFactory.getPlatformMXBeans(MemoryPoolMXBean.class)) {
+    for (MemoryPoolMXBean mbean : ManagementFactory.getMemoryPoolMXBeans()) {
       registry.register(new MemoryPoolMeter(registry, mbean));
     }
     for (BufferPoolMXBean mbean : ManagementFactory.getPlatformMXBeans(BufferPoolMXBean.class)) {
@@ -53,7 +53,7 @@ public final class Jmx {
   }
 
   private static void monitorClassLoadingMXBean(Registry registry) {
-    ClassLoadingMXBean classLoadingMXBean = ManagementFactory.getPlatformMXBean(ClassLoadingMXBean.class);
+    ClassLoadingMXBean classLoadingMXBean = ManagementFactory.getClassLoadingMXBean();
     PolledMeter.using(registry)
       .withName("jvm.classloading.classesLoaded")
       .monitorMonotonicCounter(classLoadingMXBean, ClassLoadingMXBean::getTotalLoadedClassCount);
@@ -63,7 +63,7 @@ public final class Jmx {
   }
 
   private static void monitorThreadMXBean(Registry registry) {
-    ThreadMXBean threadMXBean = ManagementFactory.getPlatformMXBean(ThreadMXBean.class);
+    ThreadMXBean threadMXBean = ManagementFactory.getThreadMXBean();
     PolledMeter.using(registry)
       .withName("jvm.thread.startedThreadCount")
       .monitorMonotonicCounter(threadMXBean, ThreadMXBean::getTotalStartedThreadCount);
@@ -76,7 +76,7 @@ public final class Jmx {
   }
 
   private static void monitorCompilationMXBean(Registry registry) {
-    CompilationMXBean compilationMXBean = ManagementFactory.getPlatformMXBean(CompilationMXBean.class);
+    CompilationMXBean compilationMXBean = ManagementFactory.getCompilationMXBean();
     if (compilationMXBean.isCompilationTimeMonitoringSupported()) {
       PolledMeter.using(registry)
         .withName("jvm.compilation.compilationTime")
--- a/spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
+++ b/spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
@@ -15,6 +15,7 @@
  */
 package com.netflix.spectator.jvm;
 
+import com.netflix.spectator.api.Gauge;
 import com.netflix.spectator.api.Registry;
 import com.netflix.spectator.api.patterns.PolledMeter;
 import com.typesafe.config.Config;
@@ -65,14 +66,17 @@ public final class Jmx {
   private static void monitorThreadMXBean(Registry registry) {
     ThreadMXBean threadMXBean = ManagementFactory.getThreadMXBean();
     PolledMeter.using(registry)
-      .withName("jvm.thread.startedThreadCount")
+      .withName("jvm.thread.threadsStarted")
       .monitorMonotonicCounter(threadMXBean, ThreadMXBean::getTotalStartedThreadCount);
-    PolledMeter.using(registry)
-      .withName("jvm.thread.threadCount")
-      .monitorValue(threadMXBean, ThreadMXBean::getThreadCount);
-    PolledMeter.using(registry)
-      .withName("jvm.thread.daemonThreadCount")
-      .monitorValue(threadMXBean, ThreadMXBean::getDaemonThreadCount);
+
+    Gauge nonDaemonThreadCount = registry.gauge("jvm.thread.threadCount", "id", "non-daemon");
+    Gauge daemonThreadCount = registry.gauge("jvm.thread.threadCount", "id", "daemon");
+    PolledMeter.poll(registry, () -> {
+      int threads = threadMXBean.getThreadCount();
+      int daemonThreads = threadMXBean.getDaemonThreadCount();
+      nonDaemonThreadCount.set(Math.max(0, threads - daemonThreads));
+      daemonThreadCount.set(daemonThreads);
+    });
   }
 
   private static void monitorCompilationMXBean(Registry registry) {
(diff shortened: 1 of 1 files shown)


spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
```diff
@@ -36,15 +35,51 @@ private Jmx() {
...
+      .monitorValue(threadMXBean, ThreadMXBean::getDaemonThreadCount);
+  }
+
+  private static void monitorCompilationMXBean(Registry registry) {
+    CompilationMXBean compilationMXBean = ManagementFactory.getPlatformMXBean(CompilationMXBean.class);
+    if (compilationMXBean.isCompilationTimeMonitoringSupported()) {
+      PolledMeter.using(registry)
+        .withName("jvm.compilation.compilationTime")
+        .withTag("compiler", compilationMXBean.getName())
+        .monitorMonotonicCounter(compilationMXBean, CompilationMXBean::getTotalCompilationTime);
```
DanielThomas, line 79: It's been a while since I worked with Spectator. Do I need a dimension tag here or should it not matter terribly?

spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
```diff
@@ -36,15 +35,51 @@ private Jmx() {
...
+      .monitorMonotonicCounter(classLoadingMXBean, ClassLoadingMXBean::getUnloadedClassCount);
+  }
+
+  private static void monitorThreadMXBean(Registry registry) {
+    ThreadMXBean threadMXBean = ManagementFactory.getPlatformMXBean(ThreadMXBean.class);
+    PolledMeter.using(registry)
+      .withName("jvm.thread.startedThreadCount")
+      .monitorMonotonicCounter(threadMXBean, ThreadMXBean::getTotalStartedThreadCount);
+    PolledMeter.using(registry)
+      .withName("jvm.thread.threadCount")
```
copperlight, line 66: This could also be implemented as:

```
name=jvm.thread.count,id=[daemon,non-daemon]
```

Where non-daemon is:

```
(ThreadMXBean::getThreadCount - ThreadMXBean::getDaemonThreadCount)
```

This would allow for the sum of the metric to be the total count of live threads and offers a natural group-by.

The downside is duplicate calls to the `ThreadMXBean::getThreadCount` method and possible timing issues related to collecting data, if they are polled separately.

spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
```diff
@@ -16,11 +16,10 @@
 package com.netflix.spectator.jvm;
 
 import com.netflix.spectator.api.Registry;
+import com.netflix.spectator.api.patterns.PolledMeter;
 import com.typesafe.config.Config;
 
-import java.lang.management.BufferPoolMXBean;
-import java.lang.management.ManagementFactory;
-import java.lang.management.MemoryPoolMXBean;
+import java.lang.management.*;
```
brharrington, line 22: We typically try to avoid wildcard imports

spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
```diff
@@ -36,15 +35,51 @@ private Jmx() {
...
+    PolledMeter.using(registry)
+      .withName("jvm.thread.threadCount")
+      .monitorValue(threadMXBean, ThreadMXBean::getThreadCount);
+    PolledMeter.using(registry)
+      .withName("jvm.thread.daemonThreadCount")
+      .monitorValue(threadMXBean, ThreadMXBean::getDaemonThreadCount);
+  }
+
+  private static void monitorCompilationMXBean(Registry registry) {
+    CompilationMXBean compilationMXBean = ManagementFactory.getPlatformMXBean(CompilationMXBean.class);
```
brharrington, line 74: `ManagementFactory.getCompilationMXBean()`

spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
brharrington: Not sure what you mean by making it a duration, do you mean have it report as a timer? As written it would report as a counter that would be `milliseconds/second` spent compiling. The unit should be converted to seconds before updating the counter as the convention is to always prefer base units. With the interface provided I don't think it can be reported as a timer since we cannot get individual samples.

spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
```diff
@@ -36,15 +35,51 @@ private Jmx() {
...
       registry.register(new MemoryPoolMeter(registry, mbean));
     }
-
     for (BufferPoolMXBean mbean : ManagementFactory.getPlatformMXBeans(BufferPoolMXBean.class)) {
       registry.register(new BufferPoolMeter(registry, mbean));
     }
   }
 
+  private static void monitorClassLoadingMXBean(Registry registry) {
+    ClassLoadingMXBean classLoadingMXBean = ManagementFactory.getPlatformMXBean(ClassLoadingMXBean.class);
```
brharrington, line 51: `ManagementFactory.getClassLoadingMXBean()`

brharrington (commented):
These metrics were collected before in the deprecated servo library:

https://github.com/Netflix/servo/blob/e29abcf27d112875fa7c7869ffa9bc2cff7d596e/servo-core/src/main/java/com/netflix/servo/publish/JvmMetricPoller.java#L47

IIRC they were omitted because they didn't get much use or created other problems. In particular, the thread related data I think was suspected to be the cause of some JVM crashes experienced on some applications. It has been a while since that came up, will try and find the old issues.

spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
DanielThomas: I had the same thought myself, and the separate polling was why I didn't. Is there a way of doing a single measurement/poll for those?

spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
DanielThomas: Alright, will make those explicit.

spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
DanielThomas: Oh, I missed there were convenience methods. I'll update the existing memory/buffer references too.

spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
DanielThomas: Oh right, yeah, timer is what I was thinking, and I'd forgotten how they're sampled.

I assume you mean something like:
```
      AtomicLong totalTime = new AtomicLong();
      PolledMeter.using(registry)
        .withName("jvm.compilation.compilationTime")
        .withTag("compiler", compilationMXBean.getName())
        .monitorValue(compilationMXBean, compilation -> {
          long currentTotalTime = compilation.getTotalCompilationTime();
          double compilationTime = (currentTotalTime - totalTime.get()) / 1000.0;
          totalTime.set(currentTotalTime);
          return compilationTime;
        });
```

brharrington: That will overwrite the gauge value with the last delta. I think the best way of handling this would be a monotonic counter that updates based on a double, see #956. Then it would just be:

```java
PolledMeter.using(registry)
        .withName("jvm.compilation.compilationTime")
        .withTag("compiler", compilationMXBean.getName())
        .monitorMonotonicCounterDouble(compilationMXBean, c -> c.getTotalCompilationTime() / 1000.0);
```

DanielThomas: Ah perfect, I'll rebase and switch back to monotonic counters.

spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
brharrington: You could do something like:

```java
    Gauge nonDaemonThreadCount = registry.gauge("jvm.thread.threadCount", "id", "non-daemon");
    Gauge daemonThreadCount = registry.gauge("jvm.thread.threadCount", "id", "daemon");
    PolledMeter.poll(registry, () -> {
      int threads = threadMXBean.getThreadCount();
      int daemonThreads = threadMXBean.getDaemonThreadCount();
      nonDaemonThreadCount.set(Math.max(0, threads - daemonThreads));
      daemonThreadCount.set(daemonThreads);
    });
```

In theory the values could change between the two calls to get the counts, but should be fairly rare and typically the rough number of threads is mainly what this is indicating.

brharrington:
Looking back at some of the old servo stuff, it would scan all the threads to collect information about number of threads in various states (running, blocked, etc). The occasional crashes seemed to be related to those scans, but we never really tracked it down so I don't know if it is fixed or not. That said, this change does not do that sort of scanning so it shouldn't be an issue.

copperlight (approved):
This is a clean implementation, with the addition of the new PolledMeter helpers. Nice work!

copperlight:
> I pushed a few minor changes. For the `jvm.thread.startedThreadCount` counter I renamed it to `jvm.thread.threadsStarted`. The goal is to avoid ambiguity with the `jvm.thread.threadCount` metric which is a gauge and I thought it could be confusing to have similar names with different types. The `threadsStarted` name lines up more closely with the counters for classloading. @copperlight can you take another look?

I like this renaming - it feels a little more consistent and better differentiates the two metric. Overall, the `jvm.thread` namespace here feels well designed.

PR #956, opened 2022-04-07
support polling monotonic double

brharrington:
Adds a helper method to PolledMeter for polling a
monotonically increasing double value. One example
where this can be useful is monotonic counters for
time that are in units like milliseconds. This allows
the value to be converted to seconds to the counter
will have a base unit.


copperlight (approved):
Nice, this makes the implementation of `.getTotalCompilationTime()` in the other PR much cleaner.

PR #957, opened 2022-04-07
helper to run task using PolledMeter pool

brharrington:
Adds a `PolledMeter.poll` method that can be used to run
an arbitrary `Runnable` task using the default thread pool
for polled meters. The main use-case is for being able to
update multiple meters each time the task is run. In this
case, the user would just use the registry to make any
updates directly.

The advantage of this method over just having the user
scheduling it themselves is it avoids having additional
polling threads running for simple use-cases. It is not
connected to the lifetime of an object, so the caller must
keep track of and use the returned future to cancel if
the polling should be stopped.

### DIFF
--- a/spectator-api/src/main/java/com/netflix/spectator/api/patterns/GaugePoller.java
+++ b/spectator-api/src/main/java/com/netflix/spectator/api/patterns/GaugePoller.java
@@ -19,6 +19,7 @@ import java.lang.ref.WeakReference;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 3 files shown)


spectator-api/src/main/java/com/netflix/spectator/api/patterns/GaugePoller.java
```diff
@@ -47,6 +48,10 @@
   private static final ScheduledExecutorService DEFAULT_EXECUTOR =
       Executors.newSingleThreadScheduledExecutor(FACTORY);
 
+  static ScheduledFuture<?> schedule(long delay, Runnable task) {
```
copperlight, line 51: I noticed in one of the other PRs that there were similar methods covering both Callable and Runnable. Should that be done here as well?

copperlight, line 51: I noticed in one of the other PRs that there were similar methods covering both Callable and Runnable. Should that be done here as well?

copperlight (commented):
Looks good, not sure if Callable should be covered in this pattern as well.

copperlight (approved):

spectator-api/src/main/java/com/netflix/spectator/api/patterns/GaugePoller.java
brharrington: No, since this is scheduling something recurring execution in the background, there would be nothing to receive the result of the callable.

PR #958, opened 2022-04-07
add registry that reports to SpectatorD

brharrington:
With this implementation Java would be consistent
with other languages based on thin clients.

### DIFF
--- a/spectator-reg-sidecar/src/main/java/com/netflix/spectator/sidecar/CommonTags.java
+++ b/spectator-reg-sidecar/src/main/java/com/netflix/spectator/sidecar/CommonTags.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2014-2022 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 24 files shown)


spectator-reg-sidecar/src/main/java/com/netflix/spectator/sidecar/CommonTags.java
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * Extract common infrastructure tags from the Netflix environment variables.
+   *
+   * @param getenv
+   *     Function used to retrieve the value of an environment variable.
+   * @return
+   *     Common tags based on the environment.
+   */
+  static Map<String, String> commonTags(Function<String, String> getenv) {
```
copperlight, line 39: Confirmed that these are the two we need to support for multi-container.

copperlight (approved):
This looks good! Very exciting to see this happen for the reference Java client.

spectator-reg-sidecar/src/test/java/com/netflix/spectator/sidecar/SidecarConfigTest.java
```diff
@@ -0,0 +1,28 @@
...
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+
+public class SidecarConfigTest {
```
copperlight, line 26: I noticed this earlier, but forgot to leave a note - was this test suite left empty intentionally?

brharrington: Addressed in #967

PR #960, opened 2022-04-08
improve indexing for :in clauses

brharrington:
Smaller in clauses will now get expanded to a disjunction
or equal clauses. This allows them to get indexed using
the faster exact match path and avoid being lumped into
the other checks. Looking at existing expressions ~95%
of in clauses are covered with the current size limit.

### DIFF
--- a/spectator-reg-atlas/src/main/java/com/netflix/spectator/atlas/impl/Query.java
+++ b/spectator-reg-atlas/src/main/java/com/netflix/spectator/atlas/impl/Query.java
@@ -587,6 +587,22 @@ public interface Query {
       return value != null && vs.contains(value);
     }
 
+    @Override public List<Query> dnfList() {
+      // For smaller sets expand to a disjunction of equal clauses. This allows them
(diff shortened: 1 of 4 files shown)


manolama (approved):

spectator-reg-atlas/src/main/java/com/netflix/spectator/atlas/impl/Query.java
```diff
@@ -587,6 +587,22 @@ public String value() {
       return value != null && vs.contains(value);
     }
 
+    @Override public List<Query> dnfList() {
+      // For smaller sets expand to a disjunction of equal clauses. This allows them
+      // to be indexed more efficiently. The size is limited because if there are
+      // multiple large in clauses in an expression the cross produce can become really
```
copperlight, line 593: s/produce/product/g

brharrington: Fixed in #963

PR #966, opened 2022-04-09
improve indexing for :has and :re

brharrington:
Two improvements to make it more efficient to check
incoming ids:

1. The :has queries are no longer lumped into the other
   checks as all that needs to be verified is that the
   key is present.
2. Uses a prefix tree for the other checks. The pattern
   matcher for :re queries can be used to extract an
   exact match prefix if possible for a regex. This makes
   it much faster to perform the matching if there are a
   lot of prefix regex queries. Looking at the current
   set of expressions internally, this is quite common.

Running a test with an index of ~341k actual expressions
and a real set of ids for incoming datapoints, the indexing
improvements significantly increase the throughput in terms
of datapoints per second checked via the index.

| Test                         | Throughput |
|------------------------------|------------|
| Baseline                     |     71,719 |
| After `:in` expansion (#960) |     97,209 |
| After this change            |    184,856 |

This change also drops an inefficient remove method from the
index to simplify ongoing maintenance.

### DIFF
--- a/spectator-reg-atlas/src/main/java/com/netflix/spectator/atlas/impl/PrefixTree.java
+++ b/spectator-reg-atlas/src/main/java/com/netflix/spectator/atlas/impl/PrefixTree.java
@@ -1,0 +1,206 @@
+/*
+ * Copyright 2014-2022 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spectator.atlas.impl;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicReferenceArray;
+import java.util.function.Consumer;
+
+/**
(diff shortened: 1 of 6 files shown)


spectator-reg-atlas/src/main/java/com/netflix/spectator/atlas/impl/PrefixTree.java
```diff
@@ -0,0 +1,206 @@
...
+        return values.remove(value);
+      } else {
+        PrefixTree<T> child = children.get(i);
+        if (child == null) {
+          return false;
+        } else {
+          boolean result = child.remove(prefix, pos + 1, value);
+          if (result && child.isEmpty()) {
+            synchronized (this) {
+              if (child.isEmpty()) {
```
manolama, line 126: Need another `children.get(i);` to double check that the child hasn't been emptied, nulled, then replaced between L120 and L125.

spectator-reg-atlas/src/main/java/com/netflix/spectator/atlas/impl/QueryIndex.java
```diff
@@ -230,7 +244,19 @@ private boolean remove(List<Query.KeyQuery> queries, int i, T value) {
...
+        } else if (kq instanceof Query.Has) {
+          if (hasKeyIdx != null) {
+            result |= hasKeyIdx.remove(queries, j, value);
+            if (hasKeyIdx.isEmpty())
+              hasKeyIdx = null;
+          }
         } else {
+          if (kq instanceof Query.Regex) {
+            Query.Regex re = (Query.Regex) kq;
+            otherChecksTree.remove(re.pattern().prefix(), kq);
```
manolama, line 256: Need `result |= ` on this and L258?

manolama (changes requested):

spectator-reg-atlas/src/main/java/com/netflix/spectator/atlas/impl/QueryIndex.java
brharrington: It shouldn't be needed, the result is updated based on the removal from the otherChecks map.

manolama (approved):

PR #969, opened 2022-04-14
add zero or one matcher

brharrington:
Before it would use an OR matcher to handle the zero or
one case. This is mostly to make it easier to experiment
with rewriting to simpler pattern formats used by some
data stores. It can be convenient to expand OR clauses,
but that can be expensive if there are many different
combinations. The zero or one, `?`, operator is commonly
used and often supported with simpler patterns so having
the explicit matcher reduces the number of combinations
from expanding OR clauses.


manolama (approved):
Good I think. I'd need to dig into the `mergeNext` cases some more to make sure I understand that part.

PR #970, opened 2022-04-19
support expanding OR clauses in regex

brharrington:
Add helper to PatternMatcher for expanding OR clauses
to a set where the original pattern should match if
at least one of the expanded patterns match. This can
be a useful first step for other processing such as
indexing the query or rewriting for data stores that
support a more limited pattern matching.

### DIFF
--- a/spectator-api/src/main/java/com/netflix/spectator/impl/matcher/Matcher.java
+++ b/spectator-api/src/main/java/com/netflix/spectator/impl/matcher/Matcher.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2019 Netflix, Inc.
+ * Copyright 2014-2022 Netflix, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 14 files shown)


spectator-api/src/main/java/com/netflix/spectator/impl/matcher/Matcher.java
```diff
@@ -53,6 +55,18 @@ default PatternMatcher ignoreCase() {
     return new IgnoreCaseMatcher(m);
   }
 
+  @Override
+  default List<PatternMatcher> expandOrClauses(int max) {
+    List<Matcher> ms = PatternUtils.expandOrClauses(this, max);
+    if (ms == null)
+      return null;
+    List<PatternMatcher> results = new ArrayList<>();
```
manolama, line 63: optional - size to `ms.size()`.

spectator-api/src/main/java/com/netflix/spectator/impl/matcher/PatternUtils.java
```diff
@@ -207,6 +210,108 @@ static Matcher ignoreCase(Matcher matcher) {
...
+    } else if (matcher instanceof ZeroOrOneMatcher) {
+      ZeroOrOneMatcher m = matcher.as();
+      List<Matcher> rs = expandOrClauses(m.repeated(), max);
+      List<Matcher> ns = expandOrClauses(m.next(), max);
+      if (rs == null || ns == null) {
+        return null;
+      } else if (rs.size() == 1 && ns.size() == 1) {
+        return Collections.singletonList(matcher);
+      } else {
+        List<Matcher> results = new ArrayList<>(ns);
```
manolama, line 233: missing `rs.size() * ns.size() > max` limit check.

manolama (changes requested):
LG I believe. Just the one missing limit check.

manolama (approved):

PR #971, opened 2022-04-23
fix removal for exprs with regex prefix

brharrington:
Before it would always remove from the prefix tree even
if there were other similar expressions with the same
prefix. After that those other expressions would never
match because the sub-index would no longer be in the
tree. Now it will only remove from the prefix tree if
the sub-index is empty.


manolama (approved):
Ah got it, sorry I didn't catch it before.

PR #979, opened 2022-07-18
Add Hotspot Safepoint metrics

DanielThomas:
Add hotspot safepoint metrics to help identify when safepoints and safepoint sync times are a problem. Unfortunately this relies on a propriety API, but there's no other way of getting at this at runtime on JDK 8 through latest. JFR event streams are JDK 15 and later for instance, and this is so much simpler. Some background here:

https://mail.openjdk.org/pipermail/serviceability-dev/2021-June/038222.html

This came up in the context of issue with I/O bound GC pauses on JDK 17:

https://twitter.com/DannyThomas/status/1546758318492102657

While I was at it, I went looking for any other causes of safepoint I/O latency and issues with safepoint sync times, and I can think of several places where we could be affected not only by that, but time-to-safepoint effective pauses where we're in JNI, etc:

- https://www.evanjones.ca/jvm-mmap-pause.html
- https://blanco.io/blog/jvm-safepoint-pauses/#analyzing-safepoint-pauses

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -5,20 +5,32 @@ on: [pull_request]
 jobs:
   build:
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        java: [8, 11, 17]
     steps:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
       - name: Gradle wrapper validation
         uses: gradle/wrapper-validation-action@v1
-      - name: Set up JDK ${{ matrix.java }}
+      - name: Set up JDK 8
         uses: actions/setup-java@v3
         with:
-          java-version: ${{ matrix.java }}
+          java-version: 8
           distribution: 'zulu'
           cache: 'gradle'
+      - run: echo "JDK8=$JAVA_HOME" >> $GITHUB_ENV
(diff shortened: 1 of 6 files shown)


DanielThomas:
Looks like `--release` isn't aware of different package access between JDK versions, and you can't specify `{{--add-exports}}.

https://bugs.openjdk.org/browse/JDK-8178152

DanielThomas:
@brharrington realized this is much simpler if I we just use toolchains. Build changes moved to a separate commit before the hotspot metrics - see what you think.

brharrington:
Thanks, I have started looking at it and should be able to go through it in detail later today.

brharrington:
I refactored it a bit to model as timers: https://github.com/brharrington/spectator/commit/6eff138815e00eec5725c7adac7c0da2d4c53e1e, doesn't look like I can push to the PR branch. Anyway let me know what you think. I'm going to build a version of the jar and run it on some test instances over the weekend.

DanielThomas:
That's awesome, thanks @brharrington! Really have to work on my Spectator foo, I never seem to get this right the first time :)

brharrington:
Closing in favor of #980

PR #982, opened 2022-08-01
support simplifying regex to expression

brharrington:
Adds method to `PatternMatcher` that attempts to rewrite
the pattern to a set of simple pattern matches that can
be combined with AND, OR, and NOT to have the same matching
behavior as the original regex pattern. This can be useful
when working with data stores that have more restricted
pattern matching support such as RE2.


spectator-api/src/test/java/com/netflix/spectator/impl/matcher/Re2PatternMatcherTest.java
```diff
@@ -0,0 +1,97 @@
...
+
+    Pattern pattern = Pattern.compile("^.*(" + regex + ")", Pattern.DOTALL);
+    if (pattern.matcher(value).find()) {
+      Assertions.assertTrue(expr.matches(value), regex + " should match " + value);
+    } else {
+      Assertions.assertFalse(expr.matches(value), regex + " shouldn't match " + value);
+    }
+  }
+
+  private static class Re2Encoder implements PatternExpr.Encoder {
```
manolama, line 44: Would this move out into a plugin eventually for RE2 based stores?

manolama (approved):
LGTM.

spectator-api/src/test/java/com/netflix/spectator/impl/matcher/Re2PatternMatcherTest.java
brharrington: The encoder implementation would be specific to a datastore and potentially how you are using it (column names, indexes, etc). I don't know that there is a need to have a common shared implementation here.

PR #988, opened 2022-09-11
Delegate to ConcurrentHashMap values Spliterator in AbstractRegistry

kilink:
Override AbstractRegistry::spliterator to return the ConcurrentHashMap's values Spliterator, which has estimated size and other characteristics set.


brharrington:
In the future, please create a fork and create the PR based on a branch in your fork. We try to only have release branches here.

PR #995, opened 2022-10-04
Lazily compute hash code in BasicTag

kilink:
Use the [racy single-check idiom](http://javaagile.blogspot.com/2013/05/the-racy-single-check-idiom.html) to lazily compute and cache the hash code in BasicTag. The hashCode method of Tag / BasicTag is not called anywhere within Spectator itself, so precomputing the hash code in the constructor is now an unnecessary overhead.

### DIFF
--- a/spectator-api/src/main/java/com/netflix/spectator/api/BasicTag.java
+++ b/spectator-api/src/main/java/com/netflix/spectator/api/BasicTag.java
@@ -34,7 +34,7 @@ public final class BasicTag implements Tag {
 
   private final String key;
   private final String value;
-  private final int hc;
+  private int hc;
 
   /**
    * Construct a new instance.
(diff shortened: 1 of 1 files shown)


spectator-api/src/main/java/com/netflix/spectator/api/BasicTag.java
```diff
@@ -66,16 +65,14 @@ public String value() {
     return key.equals(other.key) && value.equals(other.value);
   }
 
-  /**
-   * This object is immutable and the hash code is precomputed in the constructor. The id object
-   * is typically created to lookup a Meter based on dynamic dimensions so we assume that it is
```
kilink, line 71: Maybe this used to be true, but DefaultId stores its tags in an `ArrayTagSet`, so the hashCode method doesn't come into the picture.

brharrington (approved):

PR #998, opened 2022-10-07
add support for contains, starts, and ends

brharrington:
Update to be consistent with recent changes to Atlas backend to support `:contains`, `:starts`, and `:ends` query operations. These get mapped to regex so it only impacts the query parsing.

See: Netflix/atlas#1470 and Netflix/atlas#1471.


manolama (approved):
Thanks! I meant to ask if I needed to add support here as well.

PR #999, opened 2022-10-14
Skip sorting in ArrayTagSet for SortedMaps

kilink:
When tags are supplied in a SortedMap, skip the sorting step. Additionally, skip deduping when tags are supplied in a Map in general.

### DIFF
--- a/spectator-api/src/main/java/com/netflix/spectator/api/ArrayTagSet.java
+++ b/spectator-api/src/main/java/com/netflix/spectator/api/ArrayTagSet.java
@@ -20,9 +20,11 @@ import com.netflix.spectator.impl.Preconditions;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.SortedMap;
 import java.util.Spliterator;
 import java.util.Spliterators;
 import java.util.concurrent.ConcurrentMap;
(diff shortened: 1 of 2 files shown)


spectator-api/src/main/java/com/netflix/spectator/api/ArrayTagSet.java
```diff
@@ -184,11 +185,12 @@ ArrayTagSet addAll(Map<String, String> ts) {
...
       for (Map.Entry<String, String> entry : ts.entrySet()) {
-        newTags[i++] = entry.getKey();
-        newTags[i++] = entry.getValue();
+        newTags[i++] = Objects.requireNonNull(entry.getKey(), "tag keys cannot be null");
+        newTags[i++] = Objects.requireNonNull(entry.getValue(), "tag values cannot be null");
       }
-      checkForNullValues(newTags);
-      return addAll(newTags, newTags.length);
+
+      boolean sorted = ts instanceof SortedMap && ((SortedMap<?, ?>) ts).comparator() == null;
```
brharrington, line 192: Should this also check if the natural ordering is explicit? 

```java
private boolean hasNaturalOrder(SortedMap<String, String> ts) {
  return ts.comparator() == null || ts.comparator() == Comparator.<String>naturalOrder();
}
```

That would cover usage like `new TreeMap<>(Comparator.naturalOrder())`, though not `new TreeMap<>(String::compareTo)`.

kilink: I thought about checking for `Comparator.naturalOrder()`, although I think it might be unusual to explicitly supply that comparator since the Javadoc for SortedMap / TreeMap indicate natural ordering is the default. I can go ahead and add a check for that anyway, but I think it's not possible to cover all scenarios (Guava collections use an explicit comparator for natural ordering that is different than `Comparator.naturalOrder()`, and `String::compareTo` like you mentioned.

kilink: Okay, I've updated the PR to check for `Comparator.naturalOrder()` as well.

brharrington (approved):

PR #1003, opened 2022-11-11
batch updater for counter, timer, and dist summary

brharrington:
Adds a helper that can be used to batch updates to the delegate meter. The main use-case for this is timers or distributions summaries that need to be updated from a single thread with a high throughput.

For counters, this can easily be achieved by the user with a local variable to track the count and then calling increment with the amount when complete or at the desired batch interval. However, for timers and distribution summaries that doesn't work because the implementation needs the raw samples. In #941 a method was added so the samples could be accumulated in an array and posted in a batch. This helps a bit, but it is cumbersome to use, incurs memory overhead, and the full set of samples have to be traversed and processed when recorded.

This change introduces batch updaters that can be customized for a given implementation. So for registries like Atlas where it is possible to accumulate the stats without the memory overhead, it can provide an optimized implementation to do so.

### DIFF
--- a/spectator-reg-atlas/src/main/java/com/netflix/spectator/atlas/AtlasMeter.java
+++ b/spectator-reg-atlas/src/main/java/com/netflix/spectator/atlas/AtlasMeter.java
@@ -22,6 +22,7 @@ import com.netflix.spectator.api.Meter;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
 
 /** Base class for core meter types used by AtlasRegistry. */
 abstract class AtlasMeter implements Meter {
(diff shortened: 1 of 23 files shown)


spectator-reg-atlas/src/main/java/com/netflix/spectator/atlas/AtlasMeter.java
```diff
@@ -38,12 +39,19 @@ abstract class AtlasMeter implements Meter {
...
+   */
+  private AtomicInteger batchUpdaterRefCount;
+
   /** Create a new instance. */
   AtlasMeter(Id id, Clock clock, long ttl) {
     this.id = id;
     this.clock = clock;
     this.ttl = ttl;
     lastUpdated = clock.wallTime();
+    batchUpdaterRefCount = new AtomicInteger(0);
```
manolama, line 54: Is it worth looking at lazy initialization to avoid the overhead for every meter? Just worried about the high cardinality use case.

manolama (approved):
Just a bit worried about the overhead of the ref counter for instances with many, many meters but should be ok.

spectator-reg-atlas/src/main/java/com/netflix/spectator/atlas/AtlasMeter.java
brharrington: I refactored it to avoid the ref count. It is a bit less self-contained as the swap types need to inspect updater and pass in a supplier so the meter can be resolved on updates in case the registry has expired them.

PR #1007, opened 2022-11-18
Add ZGC generational memory managers

DanielThomas:
Add ZGC generational memory manager labels.

genshen uses the same memory managers, but different pools and will go into a hybrid collection when necessary, so other than checking pool size changes, I'm not sure we can do anything there:

https://github.com/openjdk/shenandoah/blob/998f68b26b8d2a5178a30a6c5b596194961e3821/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp#L2928-L2935

### DIFF
--- a/spectator-ext-gc/src/main/java/com/netflix/spectator/gc/HelperFunctions.java
+++ b/spectator-ext-gc/src/main/java/com/netflix/spectator/gc/HelperFunctions.java
@@ -65,7 +65,7 @@ final class HelperFunctions {
 
   /** Returns true if memory pool name matches an old generation pool. */
   static boolean isOldGenPool(String name) {
-    return name.endsWith("Old Gen")
+    return name.contains("Old Gen")
         || name.endsWith("Tenured Gen")
         || "Shenandoah".equals(name)
         || "ZHeap".equals(name);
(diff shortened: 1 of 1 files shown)


brharrington:
With generational ZGC is there still just a single pool name of `ZHeap`? We had to do some hacks to make allocation/promotion rate tracking work with the non-generational collectors (#847). Would that logic need to be revisited?

Similarly for generational Shenandoah is the only pool name `Shenandoah`?

DanielThomas:
Oh yes indeed, found them:

- https://github.com/openjdk/zgc/blob/zgc_generational/src/hotspot/share/gc/z/zServiceability.cpp#L180-L181
- https://github.com/openjdk/shenandoah/blob/0dfb0a3eede0f8a20286ff0edf4d50349633bff5/src/hotspot/share/gc/shenandoah/shenandoahMemoryPool.cpp#L77-L104

brharrington:
It looks like it is `Shenandoah Young Gen` and `Shenandoah Old Gen` ([code](https://github.com/openjdk/shenandoah/blob/0dfb0a3eede0f8a20286ff0edf4d50349633bff5/src/hotspot/share/gc/shenandoah/shenandoahMemoryPool.cpp#L79)). So the old gen should be fine, but we'll probably need a check for `Young Gen`, right now it looks for `Eden Space`:

https://github.com/Netflix/spectator/blob/main/spectator-ext-gc/src/main/java/com/netflix/spectator/gc/HelperFunctions.java#L72

DanielThomas:
Made it a `contains` check.

brharrington (approved):

PR #1013, opened 2023-01-03
use custom replace for substitutions

brharrington:
The variable substitution when processing JMX data used the `String.replace` function. On java 8 this internally uses regex which is costly to compile. This change switches it to a custom function for replacing the values. On newer java versions the default is more efficient and doesn't use regex, but the custom version still seems to be a little better for the sample data tested.

### DIFF
--- a/spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/MappingExpr.java
+++ b/spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/MappingExpr.java
@@ -42,6 +42,7 @@ final class MappingExpr {
    *     String with values substituted in. If no matching key is found for a
    *     placeholder, then it will not be modified and left in place.
    */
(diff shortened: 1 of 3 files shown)


spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/MappingExpr.java
```diff
@@ -29,6 +29,28 @@ final class MappingExpr {
...
+   * On Java 8, the {@code String.replace} call uses a regex internally which results
+   * in a considerable overhead. This does a simple replace without using regex.
+   */
+  static String replace(String str, String target, String replacement) {
+    int pos = str.indexOf(target);
+    if (pos == -1) {
+      return str;
+    }
+
+    StringBuilder builder = new StringBuilder();
```
manolama, line 42: may be worth sizing the builder with the original string length.

manolama (approved):

PR #1015, opened 2023-01-04
add helper for common tagging

brharrington:
Adds a sub-project with a simple helper for extracting Netflix common tags from the environment. This is using the same logic as `iep-nflxenv` which will later get updated to use this new package. The goal for this change is to remove any dependency from spectator on iep libraries that will likely move to require java 17 in the near future while spectator will need to stay on java 8 for the time being.


copperlight (approved):
Clean port from the previous iteration. Nice work!

PR #1018, opened 2023-01-04
switch spark ext to sidecar registry

jzhuge:
Spark ext uses stateless registry that is deprecated according to its README.

Switch spark ext to sidecar registry in order to send metrics to the local spectatord.

### DIFF
--- a/spectator-ext-spark/src/main/java/com/netflix/spectator/spark/SpectatorConfig.java
+++ b/spectator-ext-spark/src/main/java/com/netflix/spectator/spark/SpectatorConfig.java
@@ -15,14 +15,13 @@
  */
 package com.netflix.spectator.spark;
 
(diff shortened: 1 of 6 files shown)


spectator-ext-spark/src/main/java/com/netflix/spectator/spark/SpectatorConfig.java
```diff
@@ -35,44 +34,18 @@ public String get(String k) {
     return config.hasPath(k) ? config.getString(k) : null;
   }
 
-  @Override
   public boolean enabled() {
     return config.getBoolean("enabled");
```
jzhuge, line 38: Use `return "none".equals(config.getString("output-location"))` instead?

brharrington (approved):
Note, you will need to ensure that spectatord is on the spark nodes.

PR #1036, opened 2023-03-10
reduce locking for CompositeRegistry

brharrington:
The CompositeRegistry is the implementation used for the global registry. This change reduces the locking on the methods to get a meter. This should reduce
contention if there is a lot of concurrent access
using the global registry.

Changing the set of registries should be fairly rare. It now works similar to a copy on write array list, but is done with a simple array to avoid allocations for the common case of accessing a meter with a single registry in the composite.


manolama (approved):
Nice, that should save a good chunk of processing time.

PR #1042, opened 2023-03-15
Use a BitSet in AsciiSet

kilink:
Instead of using an array of 128 booleans, use a more compact BitSet.


brharrington:
I would have to do some testing on this. It has been a while since I checked, but my recollection was that BitSet was more memory efficient, but less CPU efficient than an array of booleans for smaller sets. For this use-case we are more concerned with the CPU efficiency.

kilink:
> I would have to do some testing on this. It has been a while since I checked, but my recollection was that BitSet was more memory efficient, but less CPU efficient than an array of booleans for smaller sets. For this use-case we are more concerned with the CPU efficiency.

We'd be able to tell from running the existing JMH benchmarks, right? I can run them and compare the results.

brharrington:
I think the `StringReplace` benchmark would cover the important aspects.

brharrington:
In my tests the BitSet does appear to be slower (#1044). Will decline this change.

PR #1043, opened 2023-03-16
Add RegisteredNameOrIp cardinality limiter

fedorka:
We have tag values which may be IP addresses, which are very high cardinality and low value, or registered names (Eureka VIPs or DNS hostnames), which are lower cardinality and high value. This limiter allows us to apply different cardinality policies to each case.

### DIFF
--- a/spectator-api/src/main/java/com/netflix/spectator/api/patterns/CardinalityLimiters.java
+++ b/spectator-api/src/main/java/com/netflix/spectator/api/patterns/CardinalityLimiters.java
@@ -17,6 +17,7 @@ package com.netflix.spectator.api.patterns;
 
 import com.netflix.spectator.api.Clock;
 import com.netflix.spectator.api.Utils;
+import com.netflix.spectator.impl.PatternMatcher;
 
 import java.io.Serializable;
 import java.util.Comparator;
@@ -28,6 +29,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 /**
@@ -153,6 +155,25 @@ public final class CardinalityLimiters {
    */
   public static Function<String, String> rollup(int n) {
     return new RollupLimiter(n);
+  }
+
+  /**
+   * Restrict the cardinality independently for values which appear to be IP literals,
+   * which are likely high-cardinality and low value, and Registered names (eg: Eureka
+   * VIPs or DNS names), which are low-to-mid cardinality and high value. Values should
+   * be RFC3986 3.2.2 hosts, behavior with non-host strings is not guaranteed.
+   *
+   * @param registeredNameLimiter
+   *     The limiter applied to values which appear to be registered names.
+   * @param ipLimiter
+   *     The limiter applied to values which appear to be IP literals.
+   *
+   * @return
+   *     The result according to the the matched limiter.
+   */
+  public static Function<String, String> registeredNameOrIp(Function<String, String> registeredNameLimiter,
+                                                            Function<String, String> ipLimiter) {
+    return new RegisteredNameOrIpLimiter(registeredNameLimiter, ipLimiter);
   }
 
   private static class FirstLimiter implements Function<String, String>, Serializable {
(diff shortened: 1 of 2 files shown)


spectator-api/src/main/java/com/netflix/spectator/api/patterns/CardinalityLimiters.java
```diff
@@ -357,4 +377,37 @@ private static class RollupLimiter implements Function<String, String>, Serializ
       return "RollupLimiter(" + state + ")";
     }
   }
+
+  private static class RegisteredNameOrIpLimiter implements Function<String, String>, Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    //From RFC 3986 3.2.2, we can quickly identify IP literals (other than IPv4) from the first character.
+    private static final Predicate<String> IS_IP_LITERAL = s -> s.startsWith("[");
```
brharrington, line 386: One observation is that this matching is inconsistent with the IPv4 check in that the IPv4 will match the pattern anywhere in the string and this would only work if it was at the start. For example:

```
http://127.0.0.1:1234/foo -> is considered an IP
http://[::1]:1234/foo     -> is not because IP is not at the start
```

Should it be `s.contains("[")`?

`IS_IP_V6_LITERAL`? I find the naming to be a bit confusing as "IP" in this context does not include IPv4 addresses. I realize in the future it could be used for future IP versions, but presently it would just be 6.

fedorka: Both good catches, thanks!

The IPv4 pattern should be anchored. The intent is for this limiter to be applicable to RFC 3986 3.2.2 hosts (or the reg_name/host components for various older standards). As noted above, this code is intended to be used on the hot path so I applied a narrow scope to the input to simplify the checks. I'll note that in the javadoc.

fedorka: Actually, I think we should stick with `IS_IP_LITERAL` as that aligns with the name in the specification, even though it does not include IPv4 addresses. The predicate for IPv4 is adjacent and both predicates are scoped to a very small section of code, so I believe the cognitive overhead is low.

Additionally, the versioning in the spec is on the format of the address, not the IP version, so parsers and matcher explicitly don't necessarily require updates when future specs are issued.

For consistency, I've also updated the IPv4 pattern to fully align with the specification.

spectator-api/src/main/java/com/netflix/spectator/api/patterns/CardinalityLimiters.java
```diff
@@ -357,4 +379,39 @@ private static class RollupLimiter implements Function<String, String>, Serializ
...
+
+    //From RFC 3986 3.2.2
+    private static final String DEC_OCTET_PATTERN = "(?:\\d|[1-9]\\d|1\\d{2}|2[0-4]\\d|25[0-5])";
```
brharrington, line 391: Multiple OR clauses will be significantly slower to match than just the simple `\d{1-3}` check used before. Not sure we really care about strict spec compliance for this in particular since IPv6 is just checking that it starts with a `[`.

spectator-api/src/main/java/com/netflix/spectator/api/patterns/CardinalityLimiters.java
```diff
@@ -357,4 +379,39 @@ private static class RollupLimiter implements Function<String, String>, Serializ
...
+    private static final String IPV4_ADDRESS_PATTERN = DEC_OCTET_PATTERN + "." + DEC_OCTET_PATTERN + "."
```
brharrington, line 392: Since the `.` is not escaped here it would match any character.

spectator-api/src/test/java/com/netflix/spectator/api/patterns/CardinalityLimitersTest.java
```diff
@@ -316,6 +316,34 @@ public void rollup2() {
     Assertions.assertEquals(CardinalityLimiters.AUTO_ROLLUP, f.apply("a"));
   }
 
+  @Test
+  public void registeredNameOrIp() {
```
brharrington, line 320: Can we add a test to verify the anchoring behavior?

spectator-api/src/main/java/com/netflix/spectator/api/patterns/CardinalityLimiters.java
brharrington: If we are going to anchor both sides of v4, then it should probably be `s.startsWith("[") && s.endsWith("]")`.

spectator-api/src/main/java/com/netflix/spectator/api/patterns/CardinalityLimiters.java
fedorka: I adjusted the implementation to align with the spec mentioned in the Javadoc. I'm happy to revert to the simpler expression and do some more wordsmithing in the docs.

brharrington: I guess to me what is odd about it is the discrepancy between the IPv4 and IPv6 checks. One is trying to be really precise and strictly match based on the spec and the other is just a simple cheap check that indicates it is likely to be an IPv6 address. If we are going to strictly check it is a valid IPv4 address according to the spec, then I would argue the same should be done for IPv6.

For this particular use-case, I think a simple check that something is likely to be an IP address is good enough and in most cases the performance benefit would be preferable. For example, for the purposes of the limiter is there harm to flagging 987.0.0.1 as an IP even though strictly speaking the number is out of range?

fedorka: My intent was to leverage the spec to write lower-cost checks. A `reg-name` cannot include a "[", so assuming that the input is a 3.2.2 `host`, it suffices to check the first character. Per the spec, an out-of-range IPv4 address would be a reg-name, hence the stricter check.

brharrington: I'm less confident it will follow that spec. In particular for Eureka VIPs which if I understand correctly are just arbitrary strings. Setup a simple benchmark to compare the more precise regex to the simpler variant:

```
Benchmark                         Mode  Cnt          Score         Error   Units
hostPrecise                      thrpt    5   21563373.607 ±  264239.452   ops/s
hostSimple                       thrpt    5  261748216.141 ± 2551757.877   ops/s
ipPrecise                        thrpt    5    6223508.840 ±    9284.224   ops/s
ipSimple                         thrpt    5   16061395.232 ±   28172.942   ops/s
```

It was about 10x slower for what I would expect to be the most common case of a host.

fedorka: Thanks for taking the time to run some benchmarks. The current version of the code uses the `\d{1,3}\.` form of the matcher.

You are correct that Eureka VIPs are arbitrary strings. This has been sufficiently problematic that we've taken some steps to clean them up.

PR #1058, opened 2023-05-11
add utility for computing 64-bit hash

brharrington:
Currently based on XXH64. Allows for data to be incrementally fed to the hash accumulator to make it more convenient to use and avoid allocations creating the structure needed to pass into the hash. Though in some cases the incremental processing is slower than just passing in something like a precomputed string.

### DIFF
--- a/spectator-api/src/main/java/com/netflix/spectator/impl/UnsafeUtils.java
+++ b/spectator-api/src/main/java/com/netflix/spectator/impl/UnsafeUtils.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright 2014-2023 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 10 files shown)


spectator-api/src/main/java/com/netflix/spectator/impl/UnsafeUtils.java
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spectator.impl;
+
+import sun.misc.Unsafe; // NOPMD
```
manolama, line 18: wow, guess it's still good through JDK 20

manolama (approved):

spectator-api/src/main/java/com/netflix/spectator/impl/UnsafeUtils.java
brharrington: Yeah, there are some other options, but are kind of a pain until we can drop java 8 support.

PR #1074, opened 2023-08-28
api: Fix duplicate tag bug.

manolama:
It was possible to provide a list of `Tag`s to an `ArrayTagSet` with duplicate values. The duplicates would not be de-duped at merge time if the duplicate keys in the list were lexically less than that of existing tags. Merge now checks the last written tag value to make sure we don't write a duplicate key.


brharrington (approved):
I adjusted the fix a bit so it would have consistent behavior with the other condition and prefer the last value for a given key.

PR #1079, opened 2023-09-15
Simplify ArrayTagSet sort method

kilink:
Get rid of unnecessary branches in ArrayTagSet's insertionSort method.


brharrington:
IIRC the branch was to avoid the loops in the simple case. We would need to do some testing to verify the performance with this change is better or neutral.

brharrington:
In my testing it makes a common case of two tags about 20% slower (#1088).

PR #1080, opened 2023-09-23
Java 21

sullis:
https://blogs.oracle.com/java/post/the-arrival-of-java-21


brharrington:
For temurin jdk21 release see: https://adoptium.net/blog/2023/09/temurin21-delay/

Temurin is used because it was included by default so it would avoid having to download a JDK in some of the builds (#1045). We would need to move to something like Azul if we don't want to wait on Temurin to release their version of JDK 21.

brharrington:
I pulled in these changes and switched back to zulu in #1081

PR #1092, opened 2023-11-08
ipc: avoid exception for empty paths

brharrington:
Fix StringIndexOutOfBoundsException if sanitize is called with an empty path string. Also updates the logic for `/` to be consistent with an empty path string. Both will now get sanitized to `none` as tag values cannot be empty.


fedorka (approved):
Thanks for the fast turnaround!

PR #1108, opened 2024-01-25
ERROR vs WARN for registry/publisher errors

vshirikian:
This should help expose Atlas reporting errors more quickly to systems that harvest and expose ERRORs


brharrington:
We keep it at warn to reduce noise for users. In the past we have gotten complaints about noise and in most cases these sort of errors are things we worry about, not app teams.

vshirikian:
Ack, thanks for the context @brharrington. We'll move to using the built-in monitoring metric.

PR #1154, opened 2024-08-29
Add Java Flight Recorder support

DanielThomas:
This adds Java Flight Recorder support using [JFR Event Streaming](https://openjdk.org/jeps/349) to begin adding measures that are not available via other means. For starters this includes per-process CPU load, Virtual Threads and ZGC.

Event Streaming was added in JDK 14, so I've anchored this to the nearest LTS using a Multi-Release jar so the classes are only visible on supported releases.

I need to dig deeper, but at a glance all of the current `registerStandardMXBeans` and `GcLogger` measures could be handled by JFR events instead. Unless there are any concerns, I'll follow this up with replacements for all of these standard measures with an `jfr` prefix so we can confirm parity with the `jvm` metrics in the real world.

### DIFF
--- a/spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
+++ b/spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
@@ -27,6 +27,8 @@ import java.lang.management.CompilationMXBean;
 import java.lang.management.ManagementFactory;
 import java.lang.management.MemoryPoolMXBean;
 import java.lang.management.ThreadMXBean;
+import java.util.concurrent.Executor;
+import java.util.concurrent.Executors;
 
 /**
  * Helpers for working with JMX mbeans.
@@ -42,6 +44,15 @@ public final class Jmx {
    * mbeans from the local jvm.
    */
   public static void registerStandardMXBeans(Registry registry) {
+    if (JavaFlightRecorder.isSupported()) {
+      Executor executor = Executors.newSingleThreadExecutor(r -> {
+        Thread t = new Thread(r, "spectator-jfr");
+        t.setDaemon(true);
+        return t;
+      });
+      JavaFlightRecorder.monitorDefaultEvents(registry, executor);
+      return;
+    }
     monitorClassLoadingMXBean(registry);
     monitorThreadMXBean(registry);
     monitorCompilationMXBean(registry);
--- a/spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
+++ b/spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
@@ -52,10 +52,11 @@ public final class Jmx {
       });
       JavaFlightRecorder.monitorDefaultEvents(registry, executor);
       return;
+    } else {
+        monitorClassLoadingMXBean(registry);
+        monitorThreadMXBean(registry);
+        monitorCompilationMXBean(registry);
     }
-    monitorClassLoadingMXBean(registry);
-    monitorThreadMXBean(registry);
-    monitorCompilationMXBean(registry);
     maybeRegisterHotspotInternal(registry);
 
     for (MemoryPoolMXBean mbean : ManagementFactory.getMemoryPoolMXBeans()) {
--- a/spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
+++ b/spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
@@ -53,9 +53,9 @@ public final class Jmx {
       JavaFlightRecorder.monitorDefaultEvents(registry, executor);
       return;
     } else {
-        monitorClassLoadingMXBean(registry);
-        monitorThreadMXBean(registry);
-        monitorCompilationMXBean(registry);
+      monitorClassLoadingMXBean(registry);
+      monitorThreadMXBean(registry);
+      monitorCompilationMXBean(registry);
     }
     maybeRegisterHotspotInternal(registry);
 
--- a/spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
+++ b/spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
@@ -13,7 +13,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package com.netflix.spectator.jvm;
 
 import com.netflix.spectator.api.Registry;
@@ -105,6 +104,7 @@ public class JavaFlightRecorder {
   }
 
   private static void collectVirtualThreadEvents(Registry registry, RecordingStream rs) {
+    // 20ms threshold set to match default behavior
     consume(VirtualThreadPinned, rs, event ->
       registry.timer("jvm.vt.pinned").record(event.getDuration())
     ).withThreshold(Duration.ofMillis(20));
@@ -114,6 +114,9 @@ public class JavaFlightRecorder {
   }
 
   private static void collectGcEvents(Registry registry, RecordingStream rs) {
+    // ZGC and Shenandoah are not covered by the generic event, there is
+    // a ZGC specific event to get coverage there, right now there doesn't
+    // appear to be similar data available for Shenandoah
     Consumer<RecordedEvent> tenuringThreshold = event ->
       registry.gauge("jvm.gc.tenuringThreshold")
         .set(event.getLong("tenuringThreshold"));
--- a/spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
+++ b/spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
@@ -96,14 +96,16 @@ public class JavaFlightRecorder {
   }
 
   private static void collectThreadStatistics(Registry registry, RecordingStream rs) {
+    Gauge nonDaemonThreadCount = registry.gauge("jvm.thread.threadCount", "id", "non-daemon");
+    Gauge daemonThreadCount = registry.gauge("jvm.thread.threadCount", "id", "daemon");
     Counter threadsStarted = registry.counter("jvm.thread.threadsStarted");
     AtomicLong prevAccumulatedCount = new AtomicLong();
     consume(JavaThreadStatistics, rs, event -> {
       long activeCount = event.getLong("activeCount");
       long daemonCount = event.getLong("daemonCount");
       long nonDaemonCount = activeCount - daemonCount;
-      registry.gauge("jvm.thread.threadCount", "id", "non-daemon").set(nonDaemonCount);
-      registry.gauge("jvm.thread.threadCount", "id", "daemon").set(daemonCount);
+      nonDaemonThreadCount.set(nonDaemonCount);
+      daemonThreadCount.set(daemonCount);
       long accumulatedCount = event.getLong("accumulatedCount");
       accumulatedCount = accumulatedCount - prevAccumulatedCount.getAndSet(accumulatedCount);
       threadsStarted.increment(accumulatedCount);
@@ -126,8 +128,8 @@ public class JavaFlightRecorder {
     // ZGC and Shenandoah are not covered by the generic event, there is
     // a ZGC specific event to get coverage there, right now there doesn't
     // appear to be similar data available for Shenandoah
-      Gauge tenuringThreshold = registry.gauge("jvm.gc.tenuringThreshold");
-      Consumer<RecordedEvent> tenuringThresholdFn = event ->
+    Gauge tenuringThreshold = registry.gauge("jvm.gc.tenuringThreshold");
+    Consumer<RecordedEvent> tenuringThresholdFn = event ->
         tenuringThreshold.set(event.getLong("tenuringThreshold"));
     consume(YoungGarbageCollection, rs, tenuringThresholdFn);
     consume(ZYoungGarbageCollection, rs, tenuringThresholdFn);
--- a/spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
+++ b/spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
@@ -1,0 +1,151 @@
+/*
+ * Copyright 2014-2024 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spectator.jvm;
+
+import com.netflix.spectator.api.Registry;
+import jdk.jfr.EventSettings;
+import jdk.jfr.consumer.RecordedEvent;
+import jdk.jfr.consumer.RecordingStream;
+
+import java.time.Duration;
+import java.util.Objects;
+import java.util.concurrent.Executor;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Consumer;
+
+public class JavaFlightRecorder {
+
+  private static final String PREFIX = "jdk.";
+  private static final String ClassLoadingStatistics = PREFIX + "ClassLoadingStatistics";
+  private static final String CompilerStatistics = PREFIX + "CompilerStatistics";
+  private static final String JavaThreadStatistics = PREFIX + "JavaThreadStatistics";
+  private static final String VirtualThreadPinned = PREFIX + "VirtualThreadPinned";
+  private static final String VirtualThreadSubmitFailed = PREFIX + "VirtualThreadSubmitFailed";
+  private static final String ZAllocationStall = PREFIX + "ZAllocationStall";
+  private static final String ZYoungGarbageCollection = PREFIX + "ZYoungGarbageCollection";
+  private static final String ZOldGarbageCollection = PREFIX + "ZOldGarbageCollection";
+
+  private JavaFlightRecorder() {
+  }
+
+  public static boolean isSupported() {
+    try {
+        Class.forName("jdk.jfr.consumer.RecordingStream");
+    } catch (ClassNotFoundException e) {
+        return false;
+    }
+    return true;
+  }
+
+  public static AutoCloseable monitorDefaultEvents(Registry registry, Executor executor) {
+    if (!isSupported()) {
+      throw new UnsupportedOperationException("This JVM does not support Java Flight Recorder event streaming");
+    }
+    Objects.requireNonNull(registry);
+    Objects.requireNonNull(executor);
+    RecordingStream rs = new RecordingStream();
+    collectClassLoadingStatistics(registry, rs);
+    collectCompilerStatistics(registry, rs);
+    collectThreadStatistics(registry, rs);
+    collectVirtualThreadEvents(registry, rs);
+    collectZgcEvents(registry, rs);
+    executor.execute(rs::start);
+    return rs::close;
+  }
+
+  private static void collectClassLoadingStatistics(Registry registry, RecordingStream rs) {
+    AtomicLong prevLoadedClassCount = new AtomicLong();
+    AtomicLong prevUnloadedClassCount = new AtomicLong();
+    consume(ClassLoadingStatistics, rs, event -> {
+      long classesLoaded = event.getLong("loadedClassCount");
+      classesLoaded = classesLoaded - prevLoadedClassCount.getAndSet(classesLoaded);
+      registry.counter("jvm.classloading.classesLoaded").increment(classesLoaded);
+
+      long classesUnloaded = event.getLong("unloadedClassCount");
+      classesUnloaded = classesUnloaded - prevUnloadedClassCount.getAndSet(classesUnloaded);
+      registry.counter("jvm.classloading.classesUnloaded").increment(classesUnloaded);
+    });
+  }
+
+  private static void collectCompilerStatistics(Registry registry, RecordingStream rs) {
+    AtomicLong prevTotalTimeSpent = new AtomicLong();
+    consume(CompilerStatistics, rs, event -> {
+      long totalTimeSpent = event.getLong("totalTimeSpent");
+      totalTimeSpent = totalTimeSpent - prevTotalTimeSpent.getAndAdd(totalTimeSpent);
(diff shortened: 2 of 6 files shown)


brharrington:
> I'll follow this up with replacements for all of these standard measures with an jfr prefix so we can confirm parity with the jvm metrics in the real world.

That is fine for testing, but we do not want users to have to deal with this. They shouldn't need to care how we get the data and I don't want to create a separate copy that could start getting some use and require migration efforts.

spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  private static void collectCpuEvents(Registry registry, RecordingStream rs) {
+    consume(CPULoad, rs, event -> {
+      String pid = "unknown";
+      try {
+        pid = String.valueOf(ProcessHandle.current().pid());
+      } catch (UnsupportedOperationException ignore) {
+      }
+      registry.gauge("jvm.cpuLoad", "type", "jvmUser", "pid", pid)
```
brharrington, line 57: Process IDs are not appropriate for metric dimensions. They aren't meaningful outside the node and can have really high cardinality in the event the process is restarting frequently.

spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
```diff
@@ -0,0 +1,107 @@
...
+    RecordingStream rs = new RecordingStream();
+    collectCpuEvents(registry, rs);
+    collectVirtualThreadEvents(registry, rs);
+    collectZgcEvents(registry, rs);
+    executor.execute(rs::start);
+    return rs::close;
+  }
+
+  private static void collectCpuEvents(Registry registry, RecordingStream rs) {
+    consume(CPULoad, rs, event -> {
```
brharrington, line 51: Would need to think about this, there are already a bunch of cpu metrics and it creates confusion for users. We try to avoid adding more unless there is a really compelling reason.

spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
```diff
@@ -0,0 +1,107 @@
...
+      String pid = "unknown";
+      try {
+        pid = String.valueOf(ProcessHandle.current().pid());
+      } catch (UnsupportedOperationException ignore) {
+      }
+      registry.gauge("jvm.cpuLoad", "type", "jvmUser", "pid", pid)
+        .set(event.getFloat("jvmUser"));
```
brharrington, line 58: What are the values? In general we try to normalize all CPU measures to be a percentage to avoid users having to worry about many different representations.

spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
```diff
@@ -0,0 +1,107 @@
...
+      registry.gauge("jvm.cpuLoad", "type", "jvmSystem", "pid", pid)
+        .set(event.getFloat("jvmSystem"));
+      registry.gauge("jvm.cpuLoad", "type", "machineTotal", "pid", pid)
+        .set(event.getFloat("machineTotal"));
```
brharrington, line 62: Would this overlap with the contribution from user/system? That would make aggregation confusing, e.g. if a user doesn't break down by type.

spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
```diff
@@ -0,0 +1,107 @@
...
+import java.time.Duration;
+import java.util.Objects;
+import java.util.concurrent.Executor;
+import java.util.concurrent.ExecutorService;
+import java.util.function.Consumer;
+
+/**
+ * Helpers supporting continuous monitoring with Java Flight Recorder.
+ */
+public class JavaFlightRecorder {
```
brharrington, line 17: My understanding was that MR jars were for internals only and could/should not be used to introduce new public APIs. From the JEP ([link](https://openjdk.org/jeps/238)):

> Every version of the library should offer the same API; investigation is required to determine whether this should be strict backwards compatibility where the API is exactly the same (byte code signature equality), or whether this can be relaxed to some degree without necessarily enabling the introduction of new enhancements that would blur the notion of one unit of release. This may imply, at a minimum, that a public class present in a release-specific directory should also be present in the root, though it need not be present in an earlier release directory. The run-time system will not verify this property, but tooling can and should detect such API compatibility issues, and a library method may also be provided to perform such varification (for example on java.util.jar.JarFile).

What implications would there be to having different public APIs? Will there be consistent behavior for both classpath and modulepath usage?

spectator-ext-jvm/src/test/java17/com/netflix/spectator/jvm/JavaFlightRecorderTest.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+    executor.shutdownNow();
+
+    Assertions.assertEquals(3, registry.measurements().count());
+    List<String> distinctMeasures = registry.measurements()
+      .map(Measurement::id)
+      .map(Id::name)
+      .distinct()
+      .toList();
+    Assertions.assertEquals(1, distinctMeasures.size());
```
brharrington, line 31: How do you know a GC event or something else couldn't have happened in the 2s window? Just want to make sure the test will not be flakey.

spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
```diff
@@ -0,0 +1,107 @@
...
+  private static void collectVirtualThreadEvents(Registry registry, RecordingStream rs) {
+    consume(VirtualThreadPinned, rs, event ->
+      registry.timer("jvm.vt.pinned").record(event.getDuration())
+    ).withThreshold(Duration.ofMillis(20));
+    consume(VirtualThreadSubmitFailed, rs, event ->
+      registry.counter("jvm.vt.submitFailed").increment()
+    );
+  }
+
+  private static void collectZgcEvents(Registry registry, RecordingStream rs) {
```
brharrington, line 75: Should there be some check of whether or not ZGC is being used?

DanielThomas:
> > I'll follow this up with replacements for all of these standard measures with an jfr prefix so we can confirm parity with the jvm metrics in the real world.
> 
> That is fine for testing, but we do not want users to have to deal with this. They shouldn't need to care how we get the data and I don't want to create a separate copy that could start getting some use and require migration efforts.

Oh yeah, totally the intention that's temporary and then rename them to `jvm.*` to replace the JMX metrics, so we can switch to just registering JFR on JDK 17 and later (i.e Spring Boot 3+).

spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
DanielThomas: No, it doesn't matter if we try to enable events that don't exist.

spectator-ext-jvm/src/test/java17/com/netflix/spectator/jvm/JavaFlightRecorderTest.java
DanielThomas: We don't, but if you're concerned about the `cpuLoad` metric, I'll find a different metric to test off and not check the total count, but that we see the expected event(s).

spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
DanielThomas: I'll make the default implementation throw an `UnsupportedOperationException` and add a helper method `isSupported` to condition off.

spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
DanielThomas: Yeah, it does. My thinking about the the CPU load was being able to reason about per-process CPU load (thus the `pid` for differentiating when you have more than one VM per-host), for multitenant JVM use cases and accounting for primary process versus sidecars. But if you're concerned about the duplication, I can totally drop this for now.

spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
DanielThomas: They're percentages, 0 to 1.0f.

spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
DanielThomas: I was looking for something to differentiate multiple JVM processes, but likely moot if we feel this doesn't have utility over an above the existing CPU instrumentation.

DanielThomas:
In fact, looks like getting parity with `GcLogger` is tricky with JFR, because the events we'd need are only in G1, so I'll leave that alone. I have added parity with `Jmx.registerStandardMXBeans` so those measures can be avoided if `JavaFlightRecorder.isSupported`.

spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
brharrington: Differentiating processes can be done by setting the [NETFLIX_PROCESS_NAME](https://netflix.github.io/atlas-docs/spectator/lang/go/migrations/#common-tags) to a meaningful value. That then applies to all metrics for that process.

spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
brharrington: If it is kept, then I think we would want to normalize to 0-100 so it matches others. Also what is the time frame for that computation?

spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
brharrington: In this comment it is more a question of the modeling. Users should be able to incrementally drill into the data. So if the user just selects `name,jvm.cpuLoad,:eq,:sum`, then they would be aggregating over all the types. If `machineTotal` overlaps with the other types, then they get a confusing result. But yeah, if we just drop the CPU stuff we can ignore that.

spectator-ext-jvm/src/test/java17/com/netflix/spectator/jvm/JavaFlightRecorderTest.java
brharrington: Just checking for some events seems reasonable if we cannot guarantee the number. So long as the test is robust and doesn't spuriously fail.

brharrington:
> In fact, looks like getting parity with GcLogger is tricky with JFR, because the events we'd need are only in G1, so I'll leave that alone

Interesting, do you know if that is something that the JDK team is looking at? What are the specific gaps? I think [jdk.GarbageCollection](https://bestsolution-at.github.io/jfr-doc/openjdk-18.html#jdk.GarbageCollection) would cover the main pause times. Concurrent phase time and before and after memory pool sizes is a bit less clear from my quick skimming.

spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
```diff
@@ -0,0 +1,151 @@
...
+  private static final String ZAllocationStall = PREFIX + "ZAllocationStall";
+  private static final String ZYoungGarbageCollection = PREFIX + "ZYoungGarbageCollection";
+  private static final String ZOldGarbageCollection = PREFIX + "ZOldGarbageCollection";
+
+  private JavaFlightRecorder() {
+  }
+
+  public static boolean isSupported() {
+    try {
+        Class.forName("jdk.jfr.consumer.RecordingStream");
```
brharrington, line 47: Why would this be missing on jdk17+?

spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
```diff
@@ -0,0 +1,151 @@
...
+    consume(CompilerStatistics, rs, event -> {
+      long totalTimeSpent = event.getLong("totalTimeSpent");
+      totalTimeSpent = totalTimeSpent - prevTotalTimeSpent.getAndAdd(totalTimeSpent);
+      registry.counter("jvm.compilation.compilationTime").add(totalTimeSpent / 1000.0);
+    });
+  }
+
+  private static void collectThreadStatistics(Registry registry, RecordingStream rs) {
+    AtomicLong prevAccumulatedCount = new AtomicLong();
+    consume(JavaThreadStatistics, rs, event -> {
```
brharrington, line 95: Is there a predictable frequency to these events or would they only occur on changes? For example if threads are not getting started/stopped for a long time would there still be an event? Spectator meters will expire if there is no activity for a while, so that might be an issue.

spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
```diff
@@ -0,0 +1,151 @@
...
+      long accumulatedCount = event.getLong("accumulatedCount");
+      long threadsStarted = accumulatedCount - prevAccumulatedCount.getAndSet(accumulatedCount);
+      registry.counter("jvm.thread.threadsStarted").increment(threadsStarted);
+    });
+  }
+
+  private static void collectVirtualThreadEvents(Registry registry, RecordingStream rs) {
+    consume(VirtualThreadPinned, rs, event ->
+      registry.timer("jvm.vt.pinned").record(event.getDuration())
+    ).withThreshold(Duration.ofMillis(20));
```
brharrington, line 110: Can you add a comment on why 20ms was chosen as the threshold?

spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
```diff
@@ -0,0 +1,151 @@
...
+    consume(VirtualThreadSubmitFailed, rs, event ->
+      registry.counter("jvm.vt.submitFailed").increment()
+    );
+  }
+
+  private static void collectZgcEvents(Registry registry, RecordingStream rs) {
+    consume(ZYoungGarbageCollection, rs, event ->
+      registry.timer("jvm.zgc.youngCollection", "tenuringThreshold", event.getString("tenuringThreshold"))
```
brharrington, line 118: `tenuringThreshold` appears to be a numeric value and if I understand correctly, ZGC can dynamically adjust that value. It would probably be better to have that as a separate metric rather than a dimension.

spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
```diff
@@ -0,0 +1,151 @@
...
+        .record(event.getDuration()));
+
+    consume(ZOldGarbageCollection, rs, event ->
+      registry.timer("jvm.zgc.oldCollection")
```
brharrington, line 122: How would these timers compare to existing pause and concurrent phase time GC timers? Presumably most of this time is not stop the world for ZGC. What additional information are these providing over the existing timers?

spectator-ext-jvm/src/test/java17/com/netflix/spectator/jvm/JavaFlightRecorderTest.java
```diff
@@ -0,0 +1,72 @@
...
+  public void isSupported() {
+    assertTrue(JavaFlightRecorder.isSupported());
+  }
+
+  @Test
+  public void checkDefaultMeasures() throws Exception {
+    Registry registry = new DefaultRegistry();
+    ExecutorService executor = Executors.newSingleThreadExecutor();
+    try (var closable = JavaFlightRecorder.monitorDefaultEvents(registry, executor)) {
+      Thread.sleep(6000);
```
brharrington, line 44: Was this chosen just to be longer than the 5s period used on the `EventSettings`?

brharrington (commented):
Since `Jmx.registerStandardMXBeans` is the main entry point used by various integrations, I think we can have it choose JFR internally when available. This would allow it to automatically transition without having to update all of different uses. Might need to have some flag to force the old behavior for testing, I'll think about that.

The GcLogger is a bit trickier, but since that isn't in scope for now can worry about that later.

spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
DanielThomas: Non-OpenJDK VMs.

spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
DanielThomas: All of the `Statistics` events are emitted at the rate configured for the `period` for the event.

spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
DanielThomas: Mirroring the default:

https://github.com/openjdk/jdk/blob/62dad3a9ea222b0fbf15668d6e7b1c4ed61b2532/src/jdk.jfr/share/conf/jfr/default.jfc#L75-L79

spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
DanielThomas: Oh yeah, good point. Let me do that.

spectator-ext-jvm/src/test/java17/com/netflix/spectator/jvm/JavaFlightRecorderTest.java
DanielThomas: That's right.

spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
DanielThomas: Thinking about this some more, we're getting accurate timings for young/old, so let me just switch this to adding a `jvm.gc.tenuringThreshold` and `jvm.gc.allocationStall` rather than `jvm.zgc`. The former is available for G1 as well, and the latter can also apply to Shenandoah if they ever emit an event.

spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
DanielThomas: Ah nice. Okay, I'll remember that for next time, but for now I don't think there's an immediate need for these, so I've dropped them.

spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
DanielThomas: Removed!

spectator-ext-jvm/src/test/java17/com/netflix/spectator/jvm/JavaFlightRecorderTest.java
DanielThomas: I've rewritten that part of the test.

DanielThomas:
> > In fact, looks like getting parity with GcLogger is tricky with JFR, because the events we'd need are only in G1, so I'll leave that alone
> 
> Interesting, do you know if that is something that the JDK team is looking at? What are the specific gaps? I think [jdk.GarbageCollection](https://bestsolution-at.github.io/jfr-doc/openjdk-18.html#jdk.GarbageCollection) would cover the main pause times. Concurrent phase time and before and after memory pool sizes is a bit less clear from my quick skimming.

I'll put a ticket in my backlog to follow up with them on lack of support for these events in the other collectors.

spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
```diff
@@ -42,6 +44,15 @@ private Jmx() {
...
+    if (JavaFlightRecorder.isSupported()) {
+      Executor executor = Executors.newSingleThreadExecutor(r -> {
+        Thread t = new Thread(r, "spectator-jfr");
+        t.setDaemon(true);
+        return t;
+      });
+      JavaFlightRecorder.monitorDefaultEvents(registry, executor);
+      return;
+    }
     monitorClassLoadingMXBean(registry);
```
brharrington, line 56: Categories that are duplicated, like classloading and compilation stats, should only be done via JMX if not using JFR.

spectator-ext-jvm/src/main/java17/com/netflix/spectator/jvm/JavaFlightRecorder.java
```diff
@@ -0,0 +1,152 @@
...
+    AtomicLong prevLoadedClassCount = new AtomicLong();
+    AtomicLong prevUnloadedClassCount = new AtomicLong();
+    consume(ClassLoadingStatistics, rs, event -> {
+      long classesLoaded = event.getLong("loadedClassCount");
+      classesLoaded = classesLoaded - prevLoadedClassCount.getAndSet(classesLoaded);
+      registry.counter("jvm.classloading.classesLoaded").increment(classesLoaded);
+
+      long classesUnloaded = event.getLong("unloadedClassCount");
+      classesUnloaded = classesUnloaded - prevUnloadedClassCount.getAndSet(classesUnloaded);
+      registry.counter("jvm.classloading.classesUnloaded").increment(classesUnloaded);
```
jasonk000, line 79: IIUC, we should be able to stash these Counter instances alongside the AtomicLong(s), instead of looking them up per-poll.

(And same for all other registry lookups below).

brharrington: Yes the meter reference can be retained to avoid looking up again. I doubt it would make much difference given the expected rate of updates here, but it would avoid a little overhead.

spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
DanielThomas: Oops, yeah, meant to make that a condition. Fixed, good eye, thanks!

brharrington (approved):

PR #1155, opened 2024-08-29
[WIP] Add new IPC tags from IPC Metrics v2 spec

umairk79:
Added new IPC tags as per the V2 spec:
 - ipc.source (To be used for api metrics)
 - ipc.method
 - ipc.source.code
 - ipc.attempt.reason
 
 Open questions for discussion:
 - Should we create a new Enum for `ApiMetric` (Similar to IpcMetric)
 - Do we need validations for `IpcAttemptReason` and `IpcMethod` enums? The values present in the enum classes aren't supposed to be exhaustive hence I didn't put in a strict validation rule (As opposed to ipc.status which has a fixed set of allowed values).
  - Should we deprecate `HttpMethod` and `HttpStatus` from IpcTags?

### DIFF
--- a/spectator-ext-ipc/src/main/java/com/netflix/spectator/ipc/IpcAttemptReason.java
+++ b/spectator-ext-ipc/src/main/java/com/netflix/spectator/ipc/IpcAttemptReason.java
@@ -1,0 +1,44 @@
+/**
+ * Copyright 2024 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spectator.ipc;
+
+import com.netflix.spectator.api.Tag;
+
+public enum IpcAttemptReason implements Tag {
+
+    /**
+     * Represents the initial attempt for the request.
+     */
+    initial,
+
+    /**
+     * Represents a retry attempt for the request.
+     */
+    retry,
+
+    /**
+     * Represents a hedge attempt for the request.
+     */
+    hedge;
+
+    @Override public String key() {
+        return IpcTagKey.attemptFinal.key();
+    }
+
+    @Override public String value() {
+        return name();
+    }
+}
--- a/spectator-ext-ipc/src/main/java/com/netflix/spectator/ipc/IpcMethod.java
+++ b/spectator-ext-ipc/src/main/java/com/netflix/spectator/ipc/IpcMethod.java
@@ -1,0 +1,94 @@
+/**
+ * Copyright 2024 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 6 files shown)


spectator-ext-ipc/src/main/java/com/netflix/spectator/ipc/IpcMetric.java
```diff
@@ -49,7 +49,9 @@ public enum IpcMetric {
           IpcTagKey.status
       ),
       EnumSet.of(
+          IpcTagKey.attemptReason,
```
nicholashagen, line 52: Another idea here that we should get more feedback on is if we should :
1. create a new set of `IpcMetricV2` enums so we separate the existing tag sets from the new tag sets
2. version this library and maintain the v1 separate from v2 (probably not enough here to warrant that)
3. keep this as-is understanding that the optional tags from v1 and v2 are slightly different

spectator-ext-ipc/src/main/java/com/netflix/spectator/ipc/IpcMethod.java
```diff
@@ -0,0 +1,94 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spectator.ipc;
+
+import com.netflix.spectator.api.Tag;
+
+public enum IpcMethod implements Tag {
+
+    /**
```
nicholashagen, line 22: Now that I look at this class, I'd like to collect feedback on if we should keep these common sets of values here or if we should treat each implementation as their own sets maintained by the implementation itself.

nicholashagen:
My thoughts on open questions:

> Should we create a new Enum for ApiMetric (Similar to IpcMetric)

I would say yeah.  I also raised the question in the code comments if we should have a `IpcMetricV2` and then maybe `ApiMetricV2` to align better and keep a cleaner definition of where the two align.

> Do we need validations for IpcAttemptReason and IpcMethod enums? The values present in the enum classes aren't supposed to be exhaustive hence I didn't put in a strict validation rule (As opposed to ipc.status which has a fixed set of allowed values)

Good question. We need more understanding from others on how these are used.  I like having concrete values provided to limit inconsistency, but also, is the non-exhaustive list more confusing to knowing it can be any string?

> Should we deprecate HttpMethod and HttpStatus from IpcTags

This is tricky.  If we don't end up versioning this library, then technically the aren't deprecated.  I think if we go the route of `IpcMetricV2` enum, then the enums would provide the set of tags and that would allow both to live side-by-side and the implementation can choose which version they want to align to (v1 or v2). At that point, there would be no need to deprecate.

brharrington:
> Should we create a new Enum for ApiMetric (Similar to IpcMetric)

Probably. I think we should try to provide as much clarity for use as possible.

spectator-ext-ipc/src/main/java/com/netflix/spectator/ipc/IpcTagKey.java
```diff
@@ -32,17 +32,31 @@ public enum IpcTagKey {
...
    * for permitted values.
    */
   status("ipc.status"),
 
+  /**
+   * Dimension indicating the transport-specific code that aligns to the IPC status.
+   * The values for this are implementation specific. For example with HTTP,
+   * the status code value would be used here.
+   */
+  statusCode("ipc.status.code"),
```
brharrington, line 53: Why doesn't `ipc.status.detail` serve this purpose? I don't think we need yet another status dimension.

spectator-ext-ipc/src/main/java/com/netflix/spectator/ipc/IpcAttemptReason.java
```diff
@@ -0,0 +1,44 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spectator.ipc;
+
+import com.netflix.spectator.api.Tag;
+
+public enum IpcAttemptReason implements Tag {
```
brharrington, line 20: This seems to overlap in large part with the existing `ipc.attempt`.

brharrington:
Closing in favor of #1168

PR #1158, opened 2024-09-14
Use AtomicLongFieldUpdater in StepLong, StepDouble, and AtomicDouble

kilink:
Use AtomicLongFieldUpdater to shrink the footprint of StepLong, StepDouble, and AtomicDouble.

From a heapdump for a typical app, we see tens of thousands of instances of these classes:

StepLong: 21,751 instances
StepDouble: 29,233 instances
AtomicDouble 31,822 instances

Getting rid of the object reference to an AtomicLong shrinks the allocation footprint and avoids the extra pointer chasing. Note that StepLong and StepDouble still have a field `current` that is an AtomicLong / AtomicDouble; it can't be inlined into a volatile field without breaking backwards compatibility since it is exposed via the `getCurrent` method.

### DIFF
--- a/spectator-api/src/main/java/com/netflix/spectator/impl/AtomicDouble.java
+++ b/spectator-api/src/main/java/com/netflix/spectator/impl/AtomicDouble.java
@@ -15,7 +15,7 @@
  */
 package com.netflix.spectator.impl;
 
-import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicLongFieldUpdater;
 
 /**
  * Wrapper around AtomicLong to make working with double values easier.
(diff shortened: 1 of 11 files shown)


spectator-api/src/main/java/com/netflix/spectator/impl/AtomicDouble.java
```diff
@@ -124,4 +127,9 @@ public void max(double v) {
   @Override public double doubleValue() {
     return get();
   }
+
+  @Override
+  public String toString() {
```
kilink, line 132: Added a `toString` implementation to match the behavior of `AtomicLong` / `AtomicInteger`

brharrington:
> Note that StepLong and StepDouble still have a field current that is an AtomicLong / AtomicDouble; it can't be inlined into a volatile field without breaking backwards compatibility since it is exposed via the getCurrent method.

Since the `impl` package is for implementation details and should only be used within spectator itself, it should be ok for those to get updated. Users should always have aligned versions of spectator sub-projects and internally that is enforced via resolution rules.

kilink:
> I think this should be good. Throughput reduction seems fairly minimal. Do you mind doing a sanity check of the commit I pushed to use the updater pattern for the current field on StepLong and StepDouble?

Looks good to me.

PR #1168, opened 2024-11-11
Add new IPC tags

ayangster:
Added new IPC tags as per the V2  spec:

- ipc.source 
- ipc.method

Extended from https://github.com/Netflix/spectator/pull/1155 with amendments.


spectator-ext-ipc/src/main/java/com/netflix/spectator/ipc/IpcMetric.java
```diff
@@ -49,7 +49,9 @@ public enum IpcMetric {
           IpcTagKey.status
       ),
       EnumSet.of(
+          IpcTagKey.source,
```
umairk79, line 52: We probably need to keep "ipc.client/server.call" enums as is and create new enum for "api.client/server.call"

ayangster: I added new enums for `api.client/server.call` and `api.client/server.inflight`. I added all the tag values as `requiredDimensions`, not sure if any of them should be `optionalDimensions`. Internally we fill out all the tag values. CC @nicholashagen @brharrington

brharrington: In terms of the required dimensions, I think it is fine to require them. In general we were probably too lenient with common IPC, ideally the validation would be strict and push to be as consistent as possible.

The common IPC inflight metrics are known to be confusing, should we avoid propagating that further and omit them from the api metrics? I would argue the common IPC ones should be deprecated and phased out over time.

ayangster: I have removed the inflight metrics following internal conversations.

PR #1207, opened 2025-08-31
Make logger field static in DefaultPublisher

kilink:
Make logger field static in DefaultPublisher.

### DIFF
--- a/spectator-reg-atlas/src/main/java/com/netflix/spectator/atlas/impl/DefaultPublisher.java
+++ b/spectator-reg-atlas/src/main/java/com/netflix/spectator/atlas/impl/DefaultPublisher.java
@@ -46,7 +46,7 @@ public final class DefaultPublisher implements Publisher {
 
   private static final String CLOCK_SKEW_TIMER = "spectator.atlas.clockSkew";
 
-  private final Logger logger = LoggerFactory.getLogger(AtlasRegistry.class);
+  private static final Logger LOGGER = LoggerFactory.getLogger(AtlasRegistry.class);
 
   private final StreamHelper streamHelper = new StreamHelper();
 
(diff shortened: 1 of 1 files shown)


spectator-reg-atlas/src/main/java/com/netflix/spectator/atlas/impl/DefaultPublisher.java
```diff
@@ -46,7 +45,7 @@ public final class DefaultPublisher implements Publisher {
 
   private static final String CLOCK_SKEW_TIMER = "spectator.atlas.clockSkew";
 
-  private final Logger logger = LoggerFactory.getLogger(AtlasRegistry.class);
+  private static final Logger LOGGER = LoggerFactory.getLogger(DefaultPublisher.class);
```
brharrington, line 48: Changing to static is fine, but the logger name shouldn't be changed at this time. There are some use-cases that expect these logs to come from the registry class.

kilink: I get that, my reasoning was that it can be confusing if the log messages aren't tied to the actual location where they're being logged, but I guess it's too late now to fix it if someone is depending on it.

kilink: Okay, I've updated the PR to just make the field static / final instead.

brharrington: > my reasoning was that it can be confusing if the log messages aren't tied to the actual location where they're being logged

Since logs are queryable and can be used for alerts, we try to be careful with certain messages, in particular the notifications round validation errors as many teams look at those.

PR #1232, opened 2026-04-08
Add jvm.gc.overhead gauge for OpenJDK 26 and later

DanielThomas:
This adds a new `jvm.gc.overhead` gauge that finally gives us an accurate measurement of GC versus application CPU utilisation, rather than using `jvm.gc.pause` as a proxy, which only works for the collectors that do no work concurrently.

### DIFF
--- a/spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
+++ b/spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
@@ -20,12 +20,9 @@ import com.netflix.spectator.api.Registry;
 import com.netflix.spectator.api.patterns.PolledMeter;
 import com.typesafe.config.Config;
 
-import java.lang.management.BufferPoolMXBean;
-import java.lang.management.ClassLoadingMXBean;
-import java.lang.management.CompilationMXBean;
-import java.lang.management.ManagementFactory;
-import java.lang.management.MemoryPoolMXBean;
-import java.lang.management.ThreadMXBean;
+import java.lang.management.*;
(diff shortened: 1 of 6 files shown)


spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
```diff
@@ -64,6 +61,30 @@ public static void registerStandardMXBeans(Registry registry) {
...
+  @SuppressWarnings("PMD.EmptyCatchBlock")
+  private static double getGcOverhead() {
+    try {
+      com.sun.management.OperatingSystemMXBean os = (com.sun.management.OperatingSystemMXBean)
+              ManagementFactory.getOperatingSystemMXBean();
+      MemoryMXBean memory = ManagementFactory.getMemoryMXBean();
+      Method getTotalGcCpuTime = MemoryMXBean.class.getMethod("getTotalGcCpuTime");
+        long gcCpuTime = (long) getTotalGcCpuTime.invoke(memory);
+        long processCpuTime = os.getProcessCpuTime();
+        return processCpuTime <= 0 ? Double.NaN : (double) gcCpuTime / processCpuTime;
```
brharrington: I don't think this calculation is correct. The `getTotalGcCpuTime` and `getProcessCpuTime` methods return the total accumulated CPU time in nanoseconds, or -1. So rather than indicating current overhead for the recent interval, it is indicating cumulative overhead for the life of the process.

DanielThomas: Oh right, I completely blanked on this not handling that for me. Looks great, thanks.

PR #1249, opened 2026-06-22
propagate autoclosables in Jmx helper class

pikinier20:


spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
```diff
@@ -43,33 +47,48 @@ private Jmx() {
...
+      closeables.add(monitorCompilationMXBean(registry));
     }
-    monitorGcOverhead(registry);
+    closeables.add(monitorGcOverhead(registry));
     for (MemoryPoolMXBean mbean : ManagementFactory.getMemoryPoolMXBeans()) {
       registry.register(new MemoryPoolMeter(registry, mbean));
     }
     for (BufferPoolMXBean mbean : ManagementFactory.getPlatformMXBeans(BufferPoolMXBean.class)) {
       registry.register(new BufferPoolMeter(registry, mbean));
     }
```
pikinier20: Out of curiosity, I think I missed it somehow: how those meters are used later? When the measurements are taken?

spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
```diff
@@ -43,33 +47,48 @@ private Jmx() {
    * Add meters for the standard MXBeans provided by the jvm. This method will use
    * {@link java.lang.management.ManagementFactory#getPlatformMXBeans(Class)} to get the set of
    * mbeans from the local jvm.
+   *
+   * <p>The returned {@link AutoCloseable} can be used to stop polling and release resources.
+   * Existing callers that do not need lifecycle management can safely ignore the return value.</p>
    */
-  public static void registerStandardMXBeans(Registry registry) {
+  public static AutoCloseable registerStandardMXBeans(Registry registry) {
```
brharrington: This is not backwards compatible. It would be source-compatible, but not binary-compatible. This change would need to be done as a new method.

spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
```diff
@@ -43,33 +47,48 @@ private Jmx() {
...
-      Executor executor = Executors.newSingleThreadExecutor(r -> {
+      ExecutorService executor = Executors.newSingleThreadExecutor(r -> {
         Thread t = new Thread(r, "spectator-jfr");
         t.setDaemon(true);
         return t;
       });
-      JavaFlightRecorder.monitorDefaultEvents(registry, executor);
+      AutoCloseable jfr = JavaFlightRecorder.monitorDefaultEvents(registry, executor);
+      closeables.add(() -> {
+        jfr.close();
```
brharrington: If this throws, the executor shutdown won't happen.

spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
brharrington: This is an older pattern that should probably get converted to PolledMeter, internally that is essentially what happens.

spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
```diff
@@ -43,33 +47,48 @@ private Jmx() {
...
+    closeables.add(monitorGcOverhead(registry));
     for (MemoryPoolMXBean mbean : ManagementFactory.getMemoryPoolMXBeans()) {
       registry.register(new MemoryPoolMeter(registry, mbean));
     }
     for (BufferPoolMXBean mbean : ManagementFactory.getPlatformMXBeans(BufferPoolMXBean.class)) {
       registry.register(new BufferPoolMeter(registry, mbean));
     }
+    return () -> {
+      for (AutoCloseable c : closeables) {
+        c.close();
```
brharrington: If any of these throw, the others would get skipped.

spectator-ext-jvm/src/main/java/com/netflix/spectator/jvm/Jmx.java
brharrington: Took a look at phasing out the deprecated usage in #1251

brharrington:
Opted for #1254 instead.

PR #1252, opened 2026-06-23
PoC: refactor GC logging

pikinier20:


brharrington:
The refactoring may be useful, will dig into it more when I get time, but since the main discussed issue was being able to use a specific registry, #1253 is a smaller change to just make the registry pluggable. Can you take a look and let me know if that addresses the primary concern?

pikinier20:
> The refactoring may be useful, will dig into it more when I get time, but since the main discussed issue was being able to use a specific registry, #1253 is a smaller change to just make the registry pluggable. Can you take a look and let me know if that addresses the primary concern?

Sure, I'm on it

brharrington:
#1253 works for now, will revisit larger refactor later.

PR #1253, opened 2026-06-24
Make GcLogger registry pluggable

brharrington:
Add a GcLogger(Registry) constructor so the GC metrics can be reported to a caller-supplied registry instead of always the global registry. The no-arg constructor delegates to the global registry, preserving existing behavior for all current callers.

The metric fields were static and bound to Spectator.globalRegistry() at class load, which a constructor parameter could not override; they are now final instance fields populated from the supplied registry. This also resolves the long-standing "bad mix of static fields used from an instance" TODO in start().

The class doc spells out the per-registry invariant (multiple instances sharing a registry double-count), and start() remains a no-op on double-start. processGcEvent is now package-private so tests can drive the event-processing path with a synthetic notification. Adds GcLoggerTest covering registry routing, isolation between instances on different registries, and the start/stop lifecycle.


pikinier20:
Thanks for implementing this change. It solves the issue in my library. I'd really want to have it merged :)

PR #1254, opened 2026-06-24
Clean up PolledMeter and registry resources on close

brharrington:
Make it possible to release the background work associated with a registry, which is important for registries created for temporary use.

PolledMeter:
- Add PolledMeter.removeAll(registry) to stop and remove all polled meters registered with a registry.
- Add Builder.withCleanupAction(AutoCloseable) so resources whose lifecycle should match a meter (e.g. a custom executor) are released when the meter is cleaned up.
- Track tasks scheduled via poll(Registry, Runnable) in the registry state so they are cancelled by removeAll/close; the returned future is wrapped so cancelling it directly also drops the bookkeeping. The return type stays ScheduledFuture for backwards compatibility.
- AbstractMeterState now implements AutoCloseable; close() cancels polling and runs cleanup actions, cleanup() also removes the state entry. The future field is volatile for safe cross-thread cancellation.

Registry:
- Registry now extends AutoCloseable with a no-op default close(), so it can be used with try-with-resources without casting and without affecting existing implementations.
- AbstractRegistry.close()/reset() release resources by closing any AutoCloseable values in the state map (e.g. PolledMeter tasks) and clearing the state and meter maps, without needing to know about the helpers built on top of the registry.
- ExtendedRegistry and MicrometerRegistry delegate close() to the wrapped registry; AtlasRegistry and SidecarRegistry release polled-meter state in addition to their existing shutdown. SidecarRegistry.close() no longer declares throws IOException.

### DIFF
--- a/spectator-api/src/main/java/com/netflix/spectator/api/patterns/PolledMeter.java
+++ b/spectator-api/src/main/java/com/netflix/spectator/api/patterns/PolledMeter.java
@@ -34,9 +34,14 @@ import java.util.Iterator;
 import java.util.Map;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.Delayed;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.DoubleSupplier;
 import java.util.function.LongSupplier;
 import java.util.function.ToDoubleFunction;
@@ -84,6 +89,10 @@ import java.util.function.ToLongFunction;
 public final class PolledMeter {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(PolledMeter.class);
+
+  // Used to generate unique ids for tracking raw poll tasks in the registry state so that they
+  // can be cancelled by removeAll(Registry).
+  private static final AtomicLong POLL_TASK_SEQ = new AtomicLong();
 
   private PolledMeter() {
   }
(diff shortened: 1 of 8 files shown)


spectator-reg-sidecar/src/main/java/com/netflix/spectator/sidecar/SidecarRegistry.java
```diff
@@ -60,11 +65,27 @@ public SidecarRegistry(Clock clock, SidecarConfig config) {
   }
 
   /**
-   * Stop the scheduler reporting data.
+   * Stop the scheduler reporting data and release associated resources. Closes any
+   * {@link AutoCloseable} state values, such as {@link PolledMeter} background tasks, before
+   * clearing the state and closing the writer.
    */
-  @Override public void close() throws IOException {
```
brharrington, line 65: I think this should be ok, but is a possible compatibility concern.

brharrington:
Example usage: https://github.com/Netflix/spectator/pull/1255

spectator-api/src/main/java/com/netflix/spectator/api/patterns/PolledMeter.java
```diff
@@ -271,6 +349,9 @@ public <T> T monitorValue(T obj, ToDoubleFunction<T> f) {
       } else {
         ValueState<T> t = (ValueState<T>) c;
         t.add(obj, f);
+        if (cleanupAction != null) {
+          t.addCleanupAction(cleanupAction);
```
manolama, line 353: There is an ordering issue here (and in the `monitorMonotonicCounterDouble`)if someone runs:
```
monitorValue()
withCleanupAction()
```
Would it be worth overloading the functions and passing in the `cleanupAction` as a third parameter? 
The only thing then would be the cleanup could be called more than once if only one is needed for multiple pollers.

manolama (approved):
One edge case around the cleanup action, otherwise looks fine.

spectator-api/src/main/java/com/netflix/spectator/api/patterns/PolledMeter.java
brharrington: The `monitorValue` method doesn't return the builder, it returns the reference object used to compute the value, so you cannot chain the `withCleanupAction` after it.

manolama: Ack, then it's just a last-withCleanupAction wins for the final use, which should be fine.

PR #1268, opened 2026-07-06
atlas: fix invalid tag characters at meter creation

brharrington:
Move the replacement of characters not permitted by the storage layer from the report paths to meter creation. Previously the fix was applied independently on the publish path (MeasurementSerializer) and the LWC streaming path (IdMapper), which allowed the two to diverge: subscription matching on the streaming path ran against the raw id, so a query written against the sanitized form (e.g. ipc.endpoint,_api_v,:re) would match the persisted data but silently drop on streaming.

Add a protected AbstractRegistry.normalizeTags hook (default no-op) that runs for every id used to create or look up a meter, and override it in AtlasRegistry to replace invalid characters using the configured validTagCharacters set. Meters are now created with a valid id, so both report paths receive already-fixed ids and no longer need to fix them. This also makes ids that differ only by invalid characters collapse to a single meter/series.

- normalizeTags returns the same id instance when nothing needs fixing (uses AsciiSet.containsAll to detect without allocating) and only rebuilds via Id.unsafeCreate when a replacement is required. A null validTagCharacters short circuits with no work.
- Remove the now-redundant fixing from MeasurementSerializer, JsonUtils, DefaultPublisher, and the EvaluatorConfig idMapper; drop the vestigial fixTagString parameter from IdMapper.
- Cache normalizeTags inside idNormalizationCache for foreign ids so the fully-normalized id is cached; DefaultId keeps the type short circuit.
- AsciiSet.containsAll: scan directly instead of comparing indexOfNonMember to length twice.

### DIFF
--- a/spectator-api/src/main/java/com/netflix/spectator/impl/AsciiSet.java
+++ b/spectator-api/src/main/java/com/netflix/spectator/impl/AsciiSet.java
@@ -188,7 +188,13 @@ public final class AsciiSet implements Serializable {
    * Returns true if all characters in the string are contained within the set.
    */
   public boolean containsAll(CharSequence str) {
-    return indexOfNonMember(str) == str.length();
+    final int n = str.length();
+    for (int i = 0; i < n; ++i) {
+      if (!contains(str.charAt(i))) {
+        return false;
+      }
+    }
+    return true;
   }
 
   private int indexOfNonMember(CharSequence str) {
--- a/spectator-reg-atlas/src/main/java/com/netflix/spectator/atlas/impl/MeasurementSerializer.java
+++ b/spectator-reg-atlas/src/main/java/com/netflix/spectator/atlas/impl/MeasurementSerializer.java
@@ -22,29 +22,21 @@ import com.netflix.spectator.api.Id;
 import com.netflix.spectator.api.Measurement;
 
 import java.io.IOException;
-import java.util.function.Function;
 
 /**
- * Jackson serializer for measurements. Values will be converted to a
- * valid set as they are written out by replacing invalid characters with
- * an '_'.
+ * Jackson serializer for measurements. Tag keys and values are written out as is; ensuring they
+ * only use characters permitted by the storage layer is the responsibility of the registry when
+ * the meter is created (see {@code AtlasRegistry.normalizeTags}). Keeping the replacement in a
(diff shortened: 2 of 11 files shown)


spectator-api/src/main/java/com/netflix/spectator/impl/AsciiSet.java
```diff
@@ -188,7 +188,13 @@ public boolean contains(char c) {
...
+    final int n = str.length();
+    for (int i = 0; i < n; ++i) {
+      if (!contains(str.charAt(i))) {
+        return false;
+      }
+    }
+    return true;
   }
 
   private int indexOfNonMember(CharSequence str) {
```
manolama, line 200: Worth deleting if we aren't using it any more?

spectator-reg-atlas/src/main/java/com/netflix/spectator/atlas/impl/MeasurementSerializer.java
```diff
@@ -22,29 +22,21 @@
...
 
-  private final Function<String, String> fixTagString;
-
-  /**
-   * Create a new instance of the serializer.
-   *
-   * @param fixTagString
-   *     Function that fixes characters used for tag keys and values.
-   */
-  public MeasurementSerializer(Function<String, String> fixTagString) {
```
manolama, line 45: This will break anyone that needs different formatting for a destination, but no one is doing that internally and likely no one externally.

manolama (approved):
It makes sense but it will impact performance for the ID creation where folks aren't memoizing the ID (the majority of cases I assume). It would further impact anything that requires replacement like IPC metrics with endpoints as the rewrite will happen on each update.

spectator-api/src/main/java/com/netflix/spectator/impl/AsciiSet.java
brharrington: It is still used in other places, it was removed from containsAll to avoid the extra length check in the hot path.

spectator-reg-atlas/src/main/java/com/netflix/spectator/atlas/impl/MeasurementSerializer.java
brharrington: It is unlikely any of those would just care about string formatting. Typically they would implement the publisher.

PR #1280, opened 2026-08-28
Remove the division from StepLong/StepDouble rollCount

michaelbraun:
## Summary

Every update to a step value calls `rollCount`, which divided the current wall time by the step size to get the interval index. `step` is a non-trusted final instance field, so C2 cannot constant fold it, and this compiles to a real `idivq` on the update path of every counter, timer, gauge and distribution summary.

Cache the end of the current interval instead of its index. An in-interval update then becomes a comparison against a volatile long, and the division is paid only on an actual rollover. Boundaries are exact multiples of `step`, so comparing them orders the intervals exactly the way comparing the indices did.

No public API or behavior change. The only externally visible difference is `toString`, which used to print `lastInitPos`; it now prints the field that actually exists. Both classes are documented as internal implementation details subject to change.

## Correctness

`StepRollCountDifferentialTest` drives the new implementation and a copy of the previous one over identical timestamp sequences and requires bit-for-bit equal results. It covers every entry point (`addAndGet`, `getCurrent`, `setCurrent`, `getAndSet`, `min`, `max`, `poll`, `pollAsRate`), asserting both the current value and `timestamp()` after each operation, over step sizes of 1ms, 7ms, 13ms, 1s, 5s and 60s and five starting alignments. The sequences deliberately include exact boundaries, boundary +/- 1, multi-interval gaps and backwards clock movement.

`StepRollCountConcurrencyTest` covers the boundary compare that precedes the CAS. Without it, two threads that both observe the same boundary would both roll, and the second would reset `current` and publish the completed interval as zero. That cannot be exercised single threaded.

I checked the tests are not vacuous by injecting four bugs; all were caught:

| Injected bug | Caught by |
|---|---|
| `>=` weakened to `>` in the boundary check | differential (all step sizes) |
| `timestamp()` off by one step | differential |
| always carry `previous` forward | differential |
| drop the boundary compare before the CAS | concurrency test only |

`./gradlew :spectator-api:build` passes, including checkstyle, spotbugs and the JDK 17/25/26 test tasks.

## Benchmark

New `StepValueUpdate` JMH benchmark in `spectator-api` (the change is here, and it needs no registry). JDK 25, 5 forks x (5 x 1s warmup + 10 x 2s measurement) = 50 measurement iterations, single threaded. Errors are 99.9% confidence intervals including fork-to-fork variance. "Before" is this repo's `main` with only the benchmark file itself added, so both runs measure identical call sequences.

|---|---:|---:|---:|---|
| `stepLongAddAndGetWithClock` | 30,501,630 ± 2,954 | 33,893,242 ± 2,623 | **+11.1%** | Production shape: clock read then update, as `AtlasCounter.add` does |
| `stepDoubleAddAndGetWithClock` | 30,235,242 ± 20,295 | 33,314,284 ± 17,130 | **+10.2%** | Same, for the double valued path |
| `stepDoublePoll` | 523,071,218 ± 23,004 | 1,421,236,753 ± 4,409,376 | **+171.7%** | Step bookkeeping alone, no atomic update |
| `stepLongPoll` | 523,059,413 ± 27,033 | 1,413,357,287 ± 3,879,963 | **+170.2%** | Step bookkeeping alone, no atomic update |
| `varyingStepLongPoll` | 523,041,511 ± 42,926 | 1,107,528,228 ± 848,189 | **+111.7%** | Bookkeeping alone, timestamp the JIT cannot hoist |
| `stepLongAddAndGet` | 492,950,974 ± 316,047 | 501,408,748 ± 848,167 | +1.7% | Bookkeeping plus the atomic add |
| `stepDoubleAddAndGet` | 228,828,698 ± 15,665 | 228,837,004 ± 15,354 | +0.0% | Bookkeeping plus the CAS loop |
| `varyingStepLongAddAndGet` | 496,459,787 ± 162,386 | 471,198,933 ± 271,861 | **-5.1%** | As above, timestamp from a counter increment |
| `rollingStepLong` | 108,540,850 ± 70,225 | 101,583,257 ± 143,372 | **-6.4%** | Worst case: 1ms step, a rollover on every call |
| `rollingStepDouble` | 108,370,102 ± 330,907 | 101,761,468 ± 133,115 | **-6.1%** | Worst case: 1ms step, a rollover on every call |
| `wallTime` | 41,131,379 ± 2,581 | 41,132,156 ± 2,244 | +0.0% | Control: one wall clock read, for scale |

`*WithClock` is the shape a real update has, matching `AtlasCounter.add`: read the clock, then update. That is the number worth quoting. The `wallTime` control matching to 0.02% across the two runs indicates the machine was not drifting between them.

The disassembly confirms the mechanism directly: `idivq` appears throughout the `before` compilation of the hot loop and is entirely absent after.

### On the two apparent regressions

`rollingStepLong` and `rollingStepDouble` use a 1ms step with a timestamp advancing every call, so every single update rolls over. That is the adversarial case: the division is still paid and the boundary load is added on top. It costs ~6%, and it only applies to a meter rolling over on essentially every update. With the 5s step used in practice a rollover is one update in millions.

`varyingStepLongAddAndGet` at -5.0% is reproducible rather than noise, and it is worth being precise about because it is not a cost of this change. The division is long-latency work on a separate execution port, so on `main` it masks whatever else the loop is doing. Measured on `main`, that benchmark is flat at 492-510M ops/s whether the timestamp comes from a constant, a counter increment, or an array load (a throwaway variant used to diagnose this, not included here). With the division gone the same three land at 501M, 471M and 453M, tracking the cost of the timestamp source itself. In other words the loop with a trivial timestamp is *faster* after this change (`stepLongAddAndGet`, 501.4M vs 493.0M); what the slower variants measure is the benchmark's own timestamp generation emerging from behind the divide.

I ruled out the alternatives before settling on that: splitting `rollCount` into a fast path plus a cold `rollCountSlow` is not responsible (an unsplit build measures the same, 470.7M vs 471.6M); nor is field layout (padding `nextStepBoundary` onto its own cache line, verified in the disassembly to move it from offset `0x30` to `0x70`, changes nothing); nor loop alignment or unrolling (`-XX:OptoLoopAlignment=32` and `-XX:LoopUnrollLimit=1` leave both builds unmoved).

A real caller reads a clock, which costs far more than anything the division could mask, which is why the production-shaped benchmark gains 11%.

### DIFF
--- a/spectator-api/src/jmh/java/com/netflix/spectator/perf/StepValueUpdate.java
+++ b/spectator-api/src/jmh/java/com/netflix/spectator/perf/StepValueUpdate.java
@@ -31,16 +31,13 @@ import org.openjdk.jmh.annotations.State;
  * measurement is the step bookkeeping plus the atomic update and not a {@code currentTimeMillis}
  * call, which is more expensive than either and would hide them.</p>
  *
- * <p>Read the {@code WithClock} variants as the headline numbers. The others deliberately leave
- * the clock read out to isolate the bookkeeping, and that makes them easy to misread: the division
- * being removed here is long-latency work on a separate execution port, so on {@code main} it
- * masks whatever else the loop is doing. Remove it and that other work stops hiding, which can
- * make a synthetic loop measure slower even though strictly less work is being done. Measured on
- * {@code main} this benchmark is flat at roughly 490-510M ops/s whether the timestamp is a
- * constant, a counter increment or an array load; with the division gone the same three land at
- * 495M, 471M and 453M, tracking the cost of the timestamp source itself. A real caller reads a
- * clock, which is far more expensive than anything the division could mask, so the effect does
- * not survive contact with the production path.</p>
+ * <p>Read the {@code WithClock} variants as the headline numbers. The others leave the clock read
+ * out to isolate the bookkeeping, which makes them easy to misread: a division is long-latency
+ * work on a separate execution port, so on {@code main} it masks whatever else the loop does.
+ * Remove it and that other work stops hiding, which can make a synthetic loop measure slower
+ * while doing strictly less. On {@code main} these are flat near 490-510M ops/s whichever way the
+ * timestamp is produced; without the division they track the cost of the timestamp source itself.
+ * A real caller reads a clock, which dwarfs anything the division could mask.</p>
  *
  * <p>{@code poll} isolates the bookkeeping on its own: it is the step handling followed by a
  * volatile read, with no atomic update. {@code addAndGet} adds the atomic update that a real
--- a/spectator-api/src/jmh/java/com/netflix/spectator/perf/StepValueUpdate.java
+++ b/spectator-api/src/jmh/java/com/netflix/spectator/perf/StepValueUpdate.java
@@ -151,7 +151,6 @@ public class StepValueUpdate {
   @State(Scope.Thread)
   public static class Varying {
     StepLong stepLong;
-    StepDouble stepDouble;
     long base;
     int i;
 
@@ -160,7 +159,6 @@ public class StepValueUpdate {
       // Align to the start of the current interval so base + 0..255 stays inside it.
       base = Clock.SYSTEM.wallTime() / 5000L * 5000L;
       stepLong = new StepLong(0L, Clock.SYSTEM, 5000L);
-      stepDouble = new StepDouble(0.0, Clock.SYSTEM, 5000L);
     }
 
     long next() {
--- a/spectator-api/src/jmh/java/com/netflix/spectator/perf/StepValueUpdate.java
+++ b/spectator-api/src/jmh/java/com/netflix/spectator/perf/StepValueUpdate.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright 2014-2025 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spectator.perf;
+
+import com.netflix.spectator.api.Clock;
+import com.netflix.spectator.impl.StepDouble;
+import com.netflix.spectator.impl.StepLong;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+
+/**
+ * Measures the update path of {@link StepLong} and {@link StepDouble}, which every counter,
+ * timer, gauge and distribution summary in a step based registry goes through.
+ *
+ * <p>The timestamp is supplied by the benchmark rather than read from a clock so that the
+ * measurement is the step bookkeeping plus the atomic update and not a {@code currentTimeMillis}
+ * call, which is more expensive than either and would hide them.</p>
+ *
+ * <p>{@code poll} isolates the bookkeeping on its own: it is the step handling followed by a
+ * volatile read, with no atomic update. {@code addAndGet} adds the atomic update that a real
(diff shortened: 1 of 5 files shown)


spectator-api/src/test/java/com/netflix/spectator/impl/StepRollCountConcurrencyTest.java
```diff
@@ -0,0 +1,167 @@
...
+   * single interval: values for intervals that roll by without being polled are dropped by
+   * design and counting them would not be a conservation property of this class.
+   */
+  @Test
+  @Timeout(60)
+  public void concurrentUpdatesWithinAnIntervalAreConserved() throws Exception {
+    ManualClock clock = new ManualClock();
+    StepLong value = new StepLong(0L, clock, STEP);
+    final int perThread = 100_000;
+    final long now = 5L * STEP;
```
brharrington, line 142: This test appears to be flakey. ManualClock starts at wallTime 0, so nextStepBoundary = STEP = 10000. But every thread calls value.addAndGet(now = 5 * STEP = 50000, 1L). The very first call satisfies now >= nextStepBoundary, so the intended "stays within a single interval" setup actually opens with a rollover. With all 8 threads already released from the barrier and incrementing, a thread that loses the boundary CAS still completes its CURRENT_UPDATER.addAndGet; if that lands before the winner's getAndSet(this, init), the increment is wiped. Hence expected: <800000> but was: <799993>.

Fix is one line before the barrier release:

```java
StepLong value = new StepLong(0L, clock, STEP);
final int perThread = 100_000;
final long now = 5L * STEP;
value.addAndGet(now, 0L);   // do the rollover up front, then the test really is in-interval
```

spectator-api/src/jmh/java/com/netflix/spectator/perf/StepValueUpdate.java
```diff
@@ -0,0 +1,182 @@
...
+  /**
+   * A timestamp that changes on every call but stays inside the current interval, so the step
+   * bookkeeping cannot be hoisted out of the measurement loop and no rollover is triggered.
+   * Separates "the division is gone" from "the JIT hoisted the division" without paying for a
+   * clock read.
+   */
+  @State(Scope.Thread)
+  public static class Varying {
+    StepLong stepLong;
+    StepDouble stepDouble;
```
brharrington: It doesn't look like this is ever used, the tests below only use the stepLong.

spectator-api/src/test/java/com/netflix/spectator/impl/StepRollCountConcurrencyTest.java
michaelbraun: Confirmed and fixed in 0fd7136d — thanks, this was a real bug in the test.

I verified the premise directly rather than trusting the reasoning. Seeding a value at t=0 and then doing the first update at `5 * STEP`:

```
timestamp before any update: 0
current after seed at t=0: 42
timestamp after first update at 5*STEP: 50000
current after it: 1        <- the 42 was wiped
```

So the rollover does happen on the first call, and the "stays within a single interval" premise the javadoc claims was never established. With eight threads already released, a loser of the boundary CAS completing its `addAndGet` before the winner's `getAndSet` loses that increment, which is exactly the 799993 you saw.

I could not reproduce it here in 25 runs, which is worth stating plainly: absence of a failure on this box is not evidence the race isn't there, and the deterministic probe above settles it.

I took your fix with one change — `value.getCurrent(now)` instead of `addAndGet(now, 0L)`, since it triggers the same rollover without an update that reads as if it were meant to contribute — plus an assertion so the precondition cannot regress silently:

```java
value.getCurrent(now);
Assertions.assertEquals(now, value.timestamp(),
    "precondition: the rollover must already have happened before the threads start");
```

Deleting the rollover line now fails that assertion deterministically rather than flaking. 25/25 green after the fix.

spectator-api/src/jmh/java/com/netflix/spectator/perf/StepValueUpdate.java
michaelbraun: Correct, removed in 0fd7136d. Only `Varying.stepDouble` was dead — the `stepDouble` fields in `InInterval` and `Rolling` are still used by the `stepDouble*` and `rollingStepDouble` benchmarks, so I left those.

spectator-api/src/test/java/com/netflix/spectator/impl/StepRollCountConcurrencyTest.java
michaelbraun: This should be fixed now

michaelbraun: should be fixed now

brharrington (approved):

PR #1281, opened 2026-08-28
Replace the per-update expiry check in SwapMeter with a removal counter

michaelbraun:
## Summary

`SwapMeter.get()` runs on the update path of every meter operation and called `underlying.hasExpired()`. For `AtlasMeter` that is a wall clock read on every update, to evaluate a TTL measured in minutes.

Track meter removal in the registry with a plain counter instead, and have `get()` compare against that. `hasExpired()` itself is unchanged, so destructive callers such as `PolledMeter` still only see real expiry.

## Why removal is a separate signal from the registry version

`SwapMeter` already had a `versionSupplier` feeding `hasExpired()`. Reusing it for removal would be smaller, but it would be wrong: callers treat `hasExpired() == true` as licence to discard the meter (`PolledMeter` drops it, `measurements()` filters it out, `cleanupCachedState()` evicts it), and removal happens on every cleanup pass rather than rarely. Folding routine removals into that signal would report healthy meters as expired once per pass. `SwapMeterExpiryReportingTest.cleanupDoesNotExpireHealthyMeters` pins that down, as does `cleanupDoesNotBlankOutACompositeRegistry` — `CompositeMeter` reports expired only when all delegates do, and `Spectator.globalRegistry()` is one.

The counter is registry-global, so any one removal invalidates every outstanding wrapper rather than only the affected one. That costs each held reference one extra map lookup per cleanup pass, which runs once per step interval. `lookupCost` below is the measurable price.

## Behavior

For a meter **past its TTL but not yet swept**, the old and new triggers are equivalent rather than merely close: `lookup()` goes through `getOrCreate`, a `computeIfAbsent` that does not replace expired entries, so the old re-resolve returned the *same instance* already held. The update lands on the same object either way. `expiredButNotYetRemovedMeterKeepsSameInstance` covers this.

Once the meter is **actually swept**, the counter moves and the next call re-resolves, which is more timely than waiting for the TTL to be observed.

For registries with **no removal signal** — `CompositeRegistry`, which has no notion of meter removal and still uses the version-only constructor — the original `underlying.hasExpired()` trigger is retained, so this change is a no-op there rather than a behavior change that has to be argued from how nested wrappers survive `unwrap()`. `legacyConstructorStillResolvesOnUnderlyingExpiry` pins it; stubbing the flag to `false` makes that test fail.

The sampling order matters and is deliberate: the removal counter is read *before* the meter is resolved. Sampled afterwards, a removal racing the lookup would already be accounted for and the wrapper would stay bound to a meter that is no longer registered, silently dropping every later update. Sampling early can only cost one redundant re-resolution. `versionIsSampledBeforeTheMeterIsResolved` drives a removal in exactly that window.

No public API is added: all five `Swap*` types are package private, and `SwapMeter` is in `impl` and documented as internal.

## Test plan

- `SwapMeterExpiryReportingTest` — routine removal is not conflated with `hasExpired()`, composites are not blanked out, the sampling order is correct, and the version-only path keeps its original trigger.
- `AtlasHeldReferenceExpiryTest` — held references recover after a sweep, updates through a stale reference are not lost, sweeping one meter does not disturb others, and an expired-but-unswept meter keeps the same instance.
- `./gradlew :spectator-api:build :spectator-reg-atlas:build` passes, including checkstyle, spotbugs and the JDK 17/25/26 test tasks.

## Benchmark

`CounterIncrement`, JDK 25, 5 forks x (5 x 1s warmup + 10 x 2s measurement) = 50 measurement iterations, single threaded. Errors are 99.9% confidence intervals including fork-to-fork variance. "Before" is `main` with only the benchmark file added, so both runs measure identical call sequences.

|---|---:|---:|---:|---|
| `swapCounter` | 17,114,400 ± 9,886 | 30,046,585 ± 44,960 | **+75.6%** | Production path: `increment()` through a held reference |
| `perThread` | 17,118,439 ± 13,348 | 30,000,862 ± 2,547 | **+75.3%** | Same path, no cache-line sharing between threads |
| `timerRecord` | 15,528,856 ± 643,774 | 26,773,658 ± 45,057 | **+72.4%** | `Timer.record()` |
| `batched` | 298,847,938 ± 219,232 | 301,014,358 ± 37,411 | +0.7% | `BatchUpdater` amortising over 1000 updates |
| `wallTime` | 40,218,577 ± 5,317 | 40,221,270 ± 5,707 | +0.0% | Control: one wall clock read, for scale |
| `atlasCounter` | 30,036,643 ± 38,915 | 30,016,521 ± 27,493 | -0.1% | Control: same path with `SwapMeter` stripped off |
| `lookupCost` | 42,628,599 ± 82,656 | 41,239,923 ± 80,697 | **-3.3%** | Cost of re-resolving a meter from the registry |

`atlasCounter` bypasses `SwapMeter` entirely, so it should not move, and it does not — that is the negative control for this change. It is worth flagging that in the combined #1277 this benchmark gained 11.8%; all of that belonged to the `rollCount` change now in #1280, and splitting the two is what made each contribution attributable.

`perThread` tracking `swapCounter` to within 0.1% confirms the gain is per-call path cost and not contention on a shared counter.

`lookupCost` regresses 3.3%, and at these error bars that is a real effect rather than noise. It is the expected trade: constructing a wrapper now samples the removal counter as well and carries one more volatile field. It is paid once per held reference per cleanup pass, against a ~1.75x saving on every single update, but it is a regression and not worth hiding.

Retaining the original trigger for the version-only path costs nothing measurable on the hot paths: `swapCounter` +0.13%, `perThread` -0.09%, `timerRecord` -0.29%, all within noise.

### DIFF
--- a/spectator-api/src/main/java/com/netflix/spectator/api/AbstractRegistry.java
+++ b/spectator-api/src/main/java/com/netflix/spectator/api/AbstractRegistry.java
@@ -37,22 +37,15 @@ import java.util.function.LongSupplier;
  */
 public abstract class AbstractRegistry implements Registry, AutoCloseable {
 
-  /**
-   * This registry has no notion of a registry level version, so the signal used by
-   * {@link SwapMeter#hasExpired()} is constant, as it has always been. Meter removal is tracked
-   * separately by {@link #removals}.
-   */
+  /** Not used for this registry, always return 0. Removal is tracked by {@link #removals}. */
   private static final LongSupplier VERSION = () -> 0L;
 
   /**
-   * Incremented whenever a meter is removed from {@link #meters}. The {@link SwapMeter} types
-   * handed out by this registry use it to notice that their underlying meter is gone and needs to
-   * be resolved again, which keeps a wall clock read off the update path.
-   *
-   * <p>Deliberately not the same signal as {@code VERSION}: that one feeds {@code hasExpired()},
-   * which callers act on destructively, and removal happens on every cleanup pass rather than
-   * rarely. Bumping one counter invalidates every outstanding wrapper, not just the affected one,
-   * but that only costs an extra map lookup per held reference per cleanup pass.
+   * Incremented whenever a meter is removed from {@link #meters}, so the {@link SwapMeter} types
+   * handed out here notice their underlying meter is gone without reading the wall clock. Kept
+   * out of {@code VERSION} because that feeds {@code hasExpired()}, which callers act on
+   * destructively. One counter invalidates every outstanding wrapper rather than just the
+   * affected one, costing an extra lookup per held reference per cleanup pass.
    */
   private final AtomicLong removals = new AtomicLong();
 
@@ -255,7 +248,7 @@ public abstract class AbstractRegistry implements Registry, AutoCloseable {
   }
 
   @Override public final Counter counter(Id id) {
-    // Sample the removal counter before resolving so a removal racing with the lookup is seen.
+    // Sampled before resolving so a removal racing the lookup is seen; see SwapMeter.
     final long v = removals.get();
     Counter c = getOrCreate(id, Counter.class, NoopCounter.INSTANCE, counterFactory);
     return new SwapCounter(this, VERSION, removalSupplier, v, c.id(), c);
@@ -345,10 +338,9 @@ public abstract class AbstractRegistry implements Registry, AutoCloseable {
   }
 
   /**
-   * Wraps the iterator over the meter map so that a removal bumps {@link #removals} no matter who
-   * performs it. Sub-classes such as {@code AtlasRegistry} implement their own cleanup pass on top
-   * of {@link #iterator()}, so the bump happens here rather than relying on every caller to
-   * remember it.
+   * Bumps {@link #removals} on removal no matter who performs it. Sub-classes such as
+   * {@code AtlasRegistry} run their own cleanup pass on top of {@link #iterator()}, so the bump
+   * belongs here rather than in every caller.
    */
   private final class VersionedIterator implements Iterator<Meter> {
 
@@ -368,8 +360,7 @@ public abstract class AbstractRegistry implements Registry, AutoCloseable {
 
     @Override public void remove() {
       delegate.remove();
-      // Bump after the entry is actually gone. A reader that sees the new version is guaranteed
-      // to miss the removed entry and will resolve a fresh meter.
+      // Bump after the entry is gone, so a reader seeing the new value cannot still find it.
       removals.incrementAndGet();
     }
   }
--- a/spectator-api/src/main/java/com/netflix/spectator/api/AbstractRegistry.java
+++ b/spectator-api/src/main/java/com/netflix/spectator/api/AbstractRegistry.java
@@ -27,6 +27,7 @@ import java.util.Map;
 import java.util.Spliterator;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Function;
 import java.util.function.LongSupplier;
 
@@ -36,8 +37,26 @@ import java.util.function.LongSupplier;
  */
 public abstract class AbstractRegistry implements Registry, AutoCloseable {
 
-  /** Not used for this registry, always return 0. */
+  /**
+   * This registry has no notion of a registry level version, so the signal used by
+   * {@link SwapMeter#hasExpired()} is constant, as it has always been. Meter removal is tracked
+   * separately by {@link #removals}.
+   */
   private static final LongSupplier VERSION = () -> 0L;
+
+  /**
+   * Incremented whenever a meter is removed from {@link #meters}. The {@link SwapMeter} types
+   * handed out by this registry use it to notice that their underlying meter is gone and needs to
+   * be resolved again, which keeps a wall clock read off the update path.
+   *
+   * <p>Deliberately not the same signal as {@code VERSION}: that one feeds {@code hasExpired()},
+   * which callers act on destructively, and removal happens on every cleanup pass rather than
+   * rarely. Bumping one counter invalidates every outstanding wrapper, not just the affected one,
+   * but that only costs an extra map lookup per held reference per cleanup pass.
+   */
+  private final AtomicLong removals = new AtomicLong();
+
+  private final LongSupplier removalSupplier = removals::get;
 
   /** Logger instance for the class. */
   protected final Logger logger;
@@ -236,32 +255,38 @@ public abstract class AbstractRegistry implements Registry, AutoCloseable {
   }
 
   @Override public final Counter counter(Id id) {
+    // Sample the removal counter before resolving so a removal racing with the lookup is seen.
+    final long v = removals.get();
     Counter c = getOrCreate(id, Counter.class, NoopCounter.INSTANCE, counterFactory);
-    return new SwapCounter(this, VERSION, c.id(), c);
+    return new SwapCounter(this, VERSION, removalSupplier, v, c.id(), c);
   }
 
   @Override public final DistributionSummary distributionSummary(Id id) {
+    final long v = removals.get();
     DistributionSummary ds = getOrCreate(
         id,
         DistributionSummary.class,
         NoopDistributionSummary.INSTANCE,
         distSummaryFactory);
-    return new SwapDistributionSummary(this, VERSION, ds.id(), ds);
+    return new SwapDistributionSummary(this, VERSION, removalSupplier, v, ds.id(), ds);
   }
 
   @Override public final Timer timer(Id id) {
+    final long v = removals.get();
     Timer t = getOrCreate(id, Timer.class, NoopTimer.INSTANCE, timerFactory);
-    return new SwapTimer(this, VERSION, t.id(), t);
+    return new SwapTimer(this, VERSION, removalSupplier, v, t.id(), t);
   }
 
   @Override public final Gauge gauge(Id id) {
+    final long v = removals.get();
     Gauge g = getOrCreate(id, Gauge.class, NoopGauge.INSTANCE, gaugeFactory);
-    return new SwapGauge(this, VERSION, g.id(), g);
+    return new SwapGauge(this, VERSION, removalSupplier, v, g.id(), g);
   }
 
   @Override public final Gauge maxGauge(Id id) {
+    final long v = removals.get();
     Gauge g = getOrCreate(id, Gauge.class, NoopGauge.INSTANCE, maxGaugeFactory);
-    return new SwapMaxGauge(this, VERSION, g.id(), g);
+    return new SwapMaxGauge(this, VERSION, removalSupplier, v, g.id(), g);
   }
 
   /**
@@ -316,7 +341,37 @@ public abstract class AbstractRegistry implements Registry, AutoCloseable {
   }
 
   @Override public final Iterator<Meter> iterator() {
-    return meters.values().iterator();
+    return new VersionedIterator(meters.values().iterator());
+  }
+
+  /**
+   * Wraps the iterator over the meter map so that a removal bumps {@link #removals} no matter who
+   * performs it. Sub-classes such as {@code AtlasRegistry} implement their own cleanup pass on top
+   * of {@link #iterator()}, so the bump happens here rather than relying on every caller to
+   * remember it.
+   */
+  private final class VersionedIterator implements Iterator<Meter> {
+
+    private final Iterator<Meter> delegate;
+
+    VersionedIterator(Iterator<Meter> delegate) {
+      this.delegate = delegate;
+    }
+
+    @Override public boolean hasNext() {
+      return delegate.hasNext();
+    }
+
+    @Override public Meter next() {
+      return delegate.next();
+    }
+
+    @Override public void remove() {
+      delegate.remove();
+      // Bump after the entry is actually gone. A reader that sees the new version is guaranteed
+      // to miss the removed entry and will resolve a fresh meter.
+      removals.incrementAndGet();
+    }
   }
 
   @Override
@@ -331,7 +386,7 @@ public abstract class AbstractRegistry implements Registry, AutoCloseable {
   protected void removeExpiredMeters() {
     int total = 0;
     int expired = 0;
-    Iterator<Meter> it = meters.values().iterator();
+    Iterator<Meter> it = iterator();
     while (it.hasNext()) {
       ++total;
       Meter m = it.next();
@@ -406,5 +461,7 @@ public abstract class AbstractRegistry implements Registry, AutoCloseable {
     }
     state.clear();
     meters.clear();
+    // Clearing the map removes every meter, so any outstanding SwapMeter has to resolve again.
+    removals.incrementAndGet();
   }
 }
--- a/spectator-api/src/main/java/com/netflix/spectator/impl/SwapMeter.java
+++ b/spectator-api/src/main/java/com/netflix/spectator/impl/SwapMeter.java
@@ -136,11 +136,14 @@ public abstract class SwapMeter<T extends Meter> implements Meter {
     // observed, which would swallow a subsequent removal.
     final long resolveVersion = resolveSupplier.getAsLong();
     if (currentResolveVersion < resolveVersion) {
-      currentResolveVersion = resolveVersion;
       // Resolving also clears the staleness hasExpired() reports, since the meter just came from
       // a fresh lookup.
       currentVersion = versionSupplier.getAsLong();
       underlying = unwrap(lookup());
+      // Published last, so a concurrent caller that sees the new version also sees the meter it
+      // describes. Publishing it first lets that caller skip the resolve and keep updating the
+      // instance this lookup just replaced, silently dropping the update.
+      currentResolveVersion = resolveVersion;
     } else if (resolveOnUnderlyingExpiry && underlying.hasExpired()) {
       currentVersion = versionSupplier.getAsLong();
       underlying = unwrap(lookup());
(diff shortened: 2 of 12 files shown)


spectator-api/src/main/java/com/netflix/spectator/impl/SwapMeter.java
```diff
@@ -78,9 +123,25 @@ public void set(T meter) {
...
+   * meter an update lands on is unchanged: one past its TTL but still registered resolves back to
+   * the same instance either way, and once cleanup removes it the counter moves.</p>
+   */
   public T get() {
-    if (hasExpired()) {
+    // Sampled once: re-reading for the assignment could store a value newer than what lookup()
+    // observed, which would swallow a subsequent removal.
+    final long resolveVersion = resolveSupplier.getAsLong();
+    if (currentResolveVersion < resolveVersion) {
+      currentResolveVersion = resolveVersion;
```
brharrington: **Lost update:** `currentResolveVersion` was published before the re-resolve completed, so a concurrent update on the same wrapper skipped the re-resolve and wrote to the meter the sweep had just removed.

> **Fixed in 2992479f** — I pushed the reorder and a regression test to this branch. Leaving the analysis below for the record.

1. A cleanup pass removes the expired meter; `removals` goes to 1.
2. Thread A calls `held.increment()`. `get()` reads `resolveVersion=1`, writes `currentResolveVersion=1`, then enters `lookup()` (`normalizeId` + CHM get + wrapper allocation).
3. Thread B calls `held.increment()` inside that window. It reads `resolveVersion=1`, sees `currentResolveVersion == 1`, so this branch is false. `resolveOnUnderlyingExpiry` is `false` for every wrapper `AbstractRegistry` hands out, so the `else if` is skipped too.
4. B returns the stale `underlying` — the counter no longer in the map — and its increment is silently dropped.

This was a regression, not a pre-existing race: the old code re-tested `underlying.hasExpired()`, which stays true until the assignment, so B re-resolved and its update landed.

Verified with a deterministic repro (registry whose `newCounter` blocks on a latch, counter that reports `hasExpired()` once removed, two threads on one held reference):

| revision | registered meter | removed meter |
| --- | --- | --- |
| `main` (baa74f4c) | **2.0** | 0.0 |
| before the fix (833ef7db) | **1.0** | 1.0 |
| after the fix (2992479f) | **2.0** | 0.0 |

The fix publishes the pre-sampled value last:

```java
if (currentResolveVersion < resolveVersion) {
  currentVersion = versionSupplier.getAsLong();
  underlying = unwrap(lookup());
  currentResolveVersion = resolveVersion;
}
```

Worst case is then a redundant lookup by the concurrent thread, which is the safe direction — the same argument the 6-arg constructor's javadoc already makes for sampling early at construction time. `:spectator-api:test` and `:spectator-reg-atlas:test` are green, and the repro (`SwapMeterConcurrentResolveTest`) fails without the reorder.

No measurable cost on the hot path, as expected — the reorder is inside a branch the benchmark never takes, since the registry is never started so `removals` stays 0:

| Benchmark | Before (ops/s) | After (ops/s) | Δ |
| --- | --- | --- | --- |
| `swapCounter` | 38,807,512 ± 575,330 | 39,050,458 ± 135,677 | +0.63% |
| `atlasCounter` (control) | 39,861,050 ± 95,246 | 39,961,246 ± 129,224 | +0.25% |
| `lookupCost` | 43,945,201 ± 229,674 | 44,093,637 ± 328,722 | +0.34% |

(3 forks x 5x1s warmup + 5x1s measurement. The control moved by a similar fraction, so that spread is run-to-run drift.)

The comment above about sampling once is right, but it's a separate concern from *when* the value becomes visible.

spectator-api/src/main/java/com/netflix/spectator/impl/SwapMeter.java
```diff
@@ -78,9 +123,25 @@ public void set(T meter) {
...
+      // Resolving also clears the staleness hasExpired() reports, since the meter just came from
+      // a fresh lookup.
+      currentVersion = versionSupplier.getAsLong();
+      underlying = unwrap(lookup());
```
brharrington: `lookup()` goes through the public registry method (`registry.counter(id)`), which allocates a fresh `SwapCounter` that `unwrap()` immediately discards.

That garbage used to be produced only for meters that had actually expired. Now one removal invalidates every outstanding wrapper, so an app holding 50k references produces 50k throwaway wrappers plus 50k `normalizeId`/`normalizeTags` passes per cleanup pass. 

An internal non-wrapping resolve (`getOrCreate` directly) would avoid both.

spectator-api/src/main/java/com/netflix/spectator/impl/SwapMeter.java
```diff
@@ -35,9 +35,21 @@ public abstract class SwapMeter<T extends Meter> implements Meter {
   /** Registry used to lookup values after expiration. */
   protected final Registry registry;
 
+  // Changes when the shape of the registry changes, for example a registry being added to a
+  // composite. Feeds hasExpired().
   private final LongSupplier versionSupplier;
   private volatile long currentVersion;
```
brharrington: For every wrapper `AbstractRegistry` hands out, `versionSupplier` is the constant `VERSION` (always 0), so `currentVersion < versionSupplier.getAsLong()` is `0 < 0` forever and the assignment in the resolve branch is a wasted volatile write. Conversely `CompositeRegistry` never bumps `removals`, so there `currentResolveVersion` duplicates `currentVersion`.

No registry uses both signals, yet `SwapMeter` now carries three extra fields (`LongSupplier` ref + `volatile long` + `boolean`, ~16 bytes) on the hottest allocation path in the library — `registry.counter(id, tags).increment()` allocates one wrapper per call. The two version fields look collapsible into one.

spectator-api/src/main/java/com/netflix/spectator/impl/SwapMeter.java
```diff
@@ -35,9 +35,21 @@ public abstract class SwapMeter<T extends Meter> implements Meter {
...
 
+  // Changes when a meter is removed, so this wrapper may be holding one that is no longer
+  // registered. Feeds get() only, never hasExpired(): callers such as PolledMeter treat expiry as
+  // licence to discard the meter, and removal happens on every cleanup pass.
+  private final LongSupplier resolveSupplier;
+  private volatile long currentResolveVersion;
+
+  // Set only by the constructor with no removal signal, so registries without one, such as
+  // CompositeRegistry, keep the original trigger.
+  private final boolean resolveOnUnderlyingExpiry;
```
brharrington: This flag exists only because `CompositeRegistry` wasn't migrated, and it puts a second path through `get()` — a hot path the PR is trying to shorten.

`CompositeRegistry` already owns an `AtomicLong` that increments on add/remove (lines 95 and 115); that *is* its removal signal. Passing it as `resolveSupplier`, with the version sampled before `newCounter(id)`, would put it on the same path as `AbstractRegistry` and let you delete this boolean, one of the two version fields, and the five 9-line pass-through constructors added to `SwapCounter`/`SwapTimer`/`SwapGauge`/`SwapMaxGauge`/`SwapDistributionSummary`.

spectator-api/src/main/java/com/netflix/spectator/api/AbstractRegistry.java
```diff
@@ -316,7 +334,35 @@ protected <T extends Meter> T getOrCreate(Id id, Class<T> cls, T dflt, Function<
...
+    }
+
+    @Override public Meter next() {
+      return delegate.next();
+    }
+
+    @Override public void remove() {
+      delegate.remove();
+      // Bump after the entry is gone, so a reader seeing the new value cannot still find it.
+      removals.incrementAndGet();
```
brharrington: `removals` is bumped once per *removed meter*, not once per cleanup pass, so the cost is not the "extra lookup per held reference per cleanup pass" the field javadoc describes.

`AtlasRegistry.removeExpiredMeters()` sweeping 5,000 expired meters bumps this 5,000 times over the duration of the loop. Any held reference updated during that window re-reads a higher value on *every* update, so each `increment()` runs `get()` → `lookup()` → `registry.counter(id)` → `normalizeId` + `normalizeTags` + CHM lookup + a throwaway wrapper, instead of the wall-clock read it replaced. A counter updating at 1M/s across a 10ms sweep does ~10k full lookups rather than one.

Coalescing to a single bump at the end of a pass would restore the documented model.

spectator-api/src/main/java/com/netflix/spectator/api/AbstractRegistry.java
```diff
@@ -36,9 +37,20 @@
  */
 public abstract class AbstractRegistry implements Registry, AutoCloseable {
 
-  /** Not used for this registry, always return 0. */
+  /** Not used for this registry, always return 0. Removal is tracked by {@link #removals}. */
   private static final LongSupplier VERSION = () -> 0L;
 
+  /**
+   * Incremented whenever a meter is removed from {@link #meters}, so the {@link SwapMeter} types
```
brharrington: `{@link SwapMeter}` doesn't resolve here — `SwapMeter` is in `com.netflix.spectator.impl` and this class doesn't import it. The build won't catch it because the javadoc task sets `Xdoclint:none`, so it ships as a dead link. Use `{@code SwapMeter}`, as this same block does two lines down for `VERSION`.

Naming nit while here: one concept carries three names across the change — `VersionedIterator`, `removals`/`removalSupplier`, and `resolveVersion`/`currentResolveVersion` in `SwapMeter` (plus "the registry version" in the PR description). Picking one term end to end, and naming the iterator after what it does, would remove the need for a couple of the explanatory comments.

spectator-reg-atlas/src/jmh/java/com/netflix/spectator/atlas/CounterIncrement.java
```diff
@@ -0,0 +1,144 @@
...
+    Timer timer;
+
+    /** Id used to measure the cost of resolving a meter from the registry. */
+    Id lookupId;
+
+    @Setup
+    public void setup() {
+      // Not started: the publishing scheduler is irrelevant to the write path and would
+      // otherwise add background work that shows up as benchmark noise.
+      registry = new AtlasRegistry(Clock.SYSTEM, System::getProperty);
```
brharrington, line 65: The registry is never started and `removeExpiredMeters()` is never called, so `removals` stays 0 for the whole run and `get()` always takes the false branch. That means no benchmark here exercises the re-resolve path the change introduces — the headline +75.6% measures only the no-removal steady state, and the per-update re-resolve during a sweep is structurally invisible. `lookupCost` measures one isolated lookup, which isn't the same thing.

A variant with a background thread calling `removeExpiredMeters()` at the step interval would show the real steady-state number.

Also: this substantially overlaps `BatchUpdates` in the same package — same registry construction, and `swapCounter`/`batched`/`timerRecord` re-measure cases it already has. Only `atlasCounter`, `lookupCost`, `wallTime` and `perThread` are new; folding those four into `BatchUpdates` would avoid a second file to keep in sync. No `@TearDown` here either, unlike `BatchUpdates`.

spectator-reg-atlas/src/test/java/com/netflix/spectator/atlas/AtlasCounterTest.java
```diff
@@ -118,6 +118,37 @@ public void expiration() {
     Assertions.assertFalse(counter.hasExpired());
   }
 
+  /**
+   * Amounts that are ignored must not refresh the last modified time. That is what allows a
+   * counter that is only ever handed junk to expire and be dropped from the registry rather
+   * than being kept alive forever by calls that record nothing.
+   */
+  @Test
+  public void ignoredAmountsDoNotRefreshExpiry() {
```
brharrington, line 127: This one looks unrelated to the PR — it covers `AtlasCounter.add()`'s pre-existing `isFinite`/positive guard before `updateLastModTime()`, and `AtlasCounter` isn't modified here. It passes identically on `main`.

Not against the coverage, but landing it in a `SwapMeter` perf change means a bisect that stops here attributes an `AtlasCounter` behavior lock-in to this commit, and reverting the perf change would drop coverage unrelated to it. Probably belongs in its own commit or PR.

spectator-api/src/test/java/com/netflix/spectator/api/SwapMeterExpiryReportingTest.java
```diff
@@ -0,0 +1,197 @@
...
+
+/**
+ * The registry bumps a version when it removes a meter so that held references know to resolve
+ * again. That signal must not leak into {@code hasExpired()}: callers treat a true result as
+ * licence to discard the meter ({@code PolledMeter} drops it, {@code measurements()} filters it
+ * out, {@code cleanupCachedState()} evicts it), and a {@link CompositeMeter} — which is what
+ * {@code Spectator.globalRegistry()} is — reports expired only when all of its delegates do, so a
+ * version based answer would blank out every meter in it after the first cleanup pass.
+ */
+public class SwapMeterExpiryReportingTest {
```
brharrington, line 29: Both new test files are single-threaded, so the concurrency-sensitive part of the change — the check-then-act on `currentResolveVersion` — had no coverage. `removalCounterIsSampledBeforeTheMeterIsResolved` uses a `getOrCreate` hook on one thread, which pins the sampling order in the constructor but not the window in `get()`.

> **Addressed in 2992479f** — I pushed `SwapMeterConcurrentResolveTest`, which pins the interleaving (one thread parked inside `lookup()` while another updates the same wrapper) and fails without the ordering fix.

I did try a load-based variant as well and dropped it: a thread that enters `get()` before the removal can legitimately write to the old meter in any lock-free design, so "no updates lost under load" isn't a sound assertion. The deterministic test pins the actual invariant instead.

brharrington (changes requested):

brharrington:
Thanks for digging into this — the steady-state win is real and worth having. I pushed three commits to the branch; the first is a straight bug fix to your design, but the second two **replace** the removal-counter mechanism, so I want to lay out the reasoning rather than leave you to reverse-engineer it from the diff.

### Why the counter had to change

`removals` is a single global counter, so any removal invalidates *every* outstanding `SwapMeter`, including wrappers holding meters that are still registered. It's also bumped once per removed meter rather than once per pass, so a cleanup pass over N meters makes each of H held references re-resolve up to N times — and each re-resolve is a full `registry.counter(id)` (normalizeId + normalizeTags + map lookup + a throwaway wrapper), rather than the wall-clock read it replaced.

That didn't show up in `CounterIncrement` because the benchmark never starts the registry and never calls `removeExpiredMeters()`, so `removals` stays at 0 and every update takes the cheap path. I added `CleanupPassCost` to measure the case the change actually affects: one cleanup pass plus one update through every held reference.

Measured against `main` (µs/op, lower is better; 2 forks, SingleShotTime so the per-invocation setup is outside the timed window):

| held | removed | main | removal counter | this branch |
| --- | --- | --- | --- | --- |
| 100 | 100 | 30.5 | 61.4 (2.02x) | 37.8 (1.24x) |
| 100 | 1000 | 91.9 | 115.6 (1.26x) | 121.6 (1.32x) |
| 1000 | 100 | 119.1 | 195.8 (1.64x) | 131.5 (1.10x) |
| 1000 | 1000 | 153.0 | 209.6 (1.37x) | 192.3 (1.26x) |

So the counter version was up to 2x slower than `main` for updates to meters that hadn't expired — the cost scales with `removed x held`, and it's paid on the update path, so more traffic makes it worse.

### What replaced it

Mark the meter instead of counting removals. `Meter` gains an `isRemoved()` default that falls back to `hasExpired()`, so registries that don't opt in keep today's behaviour; `AtlasMeter` overrides it with a flag set as part of the removal, and the iterator marks each meter it removes. A wrapper then re-resolves only when its *own* meter is gone — a held reference to a live meter never re-resolves, so `removed x held` drops out of the cost model entirely. The residual cost is roughly 26ns per removed meter, paid once for that meter rather than on every update.

Steady state is unaffected — if anything slightly better, since there's no shared `AtomicLong` line to read (`CounterIncrement.swapCounter`, ops/s):

| | ops/s | vs main |
| --- | --- | --- |
| main | 23,176,139 | — |
| removal counter | 37,105,571 | +60% |
| this branch | 39,026,152 | +68% |

For scale, `atlasCounter` — the same increment with the `SwapMeter` layer stripped off — is 39,781,758 ops/s, so the wrapper now costs about 0.5ns per update.

### Two follow-on simplifications

**The publish-ordering race.** `get()` wrote `currentResolveVersion` *before* the re-resolve completed, so a second thread on the same wrapper saw the new version, skipped the resolve and kept writing to the meter the sweep had just removed — that update was silently dropped. Verified with a deterministic repro (registry whose `newCounter` blocks on a latch, two threads on one held reference): `main` and the fixed code both land 2.0 on the registered meter, the unfixed branch lands 1.0. Fixed in the first commit by publishing the version last; the later commits make it structurally impossible, since there's no version left to publish.

**The version field.** With removal handled per-meter, the only remaining use of `versionSupplier` was `CompositeRegistry`'s shape changes — every other registry passed a constant supplier and evaluated `0 < 0` on every update. That's now a `Generation` token: one volatile boolean shared by the meters created against a given shape, with `Generation.PERMANENT` for fixed-shape registries. `SwapMeter` goes from two staleness fields to one, and both signals are now the same shape of check.

That also fixes `CompositeRegistry.removeAll()`, which installed a new empty shape without bumping the version, so held references never noticed the sub-registries had gone. Two tests added for it.

One number I can't fully explain: the sweep cost still grows somewhat with `held` even though live meters shouldn't re-resolve (6.2µs -> 15.0µs at `removed=100` going from 100 to 1000 held). It's barely outside the error bars and `updateHeldReferencesOnly` shows no such effect, so I've left it as unexplained rather than chased it.

brharrington:
I broke this up into a number of smaller PRs and changed the approach a bit to reduce the costs associated with scanning the registry for publishing and removal that was introduced with the counter approach. The perf opt is in https://github.com/Netflix/spectator/pull/1290.

PR #1298, opened 2026-09-17
Store AsciiSet membership as two longs instead of boolean[128]

michaelbraun:
## Summary

Replaces `AsciiSet`'s `boolean[128]` membership array with a pair of `long` bitmasks. Whole-set
operations drop from an O(128) loop to a couple of ALU ops, and `contains`/`containsAll` go from a
bounds-checked array load to a shift and mask.

Also gets `toPattern` out of the constructor. The old code built the pattern string eagerly for
every instance, which is what made the set algebra slow — `union` was paying for a `StringBuilder`
walk over 128 characters per call. `toString` derives it on demand instead.

## Numbers

JMH `thrpt` on JDK 25, x86-64. Each row is two jmh jars built from two git states of
`AsciiSet.java`, running the same benchmark. `AsciiSetBench` at 5 forks × 12 iterations
(60 samples); `lookupCost` at 12 forks (144 samples), because it needs them — see the note below.

**Set algebra and construction** — these are where the representation change pays off. They run
while a pattern matcher is being compiled, not per lookup.

| | main | this PR | |
|---|---|---|---|
| `union` | 4.73M ops/s | 454.73M ops/s | 96x |
| `character()` | 26.00M ops/s | 1576.51M ops/s | 61x |
| `isEmpty` | 85.89M ops/s | 1592.29M ops/s | 18.5x |
| `equals` | 110.47M ops/s | 1064.15M ops/s | 9.6x |
| `fromPattern` | 5.08M ops/s, 280 B/op | 7.47M ops/s, 32 B/op | +47%, 8.8x less garbage |

**Scanning** — this is what runs on every meter lookup, via `AtlasRegistry.normalizeTags`.

| | main | this PR | |
|---|---|---|---|
| `contains` (single char) | 1125.31M ops/s | 1306.50M ops/s | +16% |
| `containsAll`, short valid value | 118.77M ops/s | 121.61M ops/s | +2.4% |
| `containsAll`, long valid name | 51.45M ops/s | 51.68M ops/s | parity |
| `containsAll`, invalid char | 77.65M ops/s | 76.82M ops/s | **−1.1%** |
| `replaceNonMembers`, nothing to fix | 49.99M ops/s | 52.23M ops/s | +4.5% |
| `replaceNonMembers`, fix near start | 38.70M ops/s | 37.71M ops/s | **−2.6%** |
| `replaceNonMembers`, fix near end | 33.56M ops/s | 30.37M ops/s | **−9.5%** |
| `CounterIncrement.lookupCost` (real meter lookup, end to end) | 43.41M ops/s | 45.74M ops/s | +5.4% |

The three regressions are real, not noise: they reproduced at the same magnitude across
independent runs at 2 and 5 forks, with non-overlapping error bars. They are the paths that scan
*past* an invalid character, where the bitset's word-select branch costs more than the old array's
L1 load, and the array is small enough to stay resident. The end-to-end lookup is still net
positive because the dominant case is a clean value that either scans to the end or exits on the
first character.

**On the fork count**: `lookupCost` reported −5.1% at 2 forks and +6.1% at 6, each with ±0.2%
error bars. JMH's interval only covers variance within a run, so with few forks it is precise and
wrong — the per-fork code layout is the thing that moves. Three runs at ≥6 forks agree on +5.4%
to +6.1%.

## Notes

`serialVersionUID` goes to 2. Serialization matches by field name, so an old stream would
otherwise deserialize into a silently empty set.

`hashCode` multiplies before folding. `Long.hashCode` xors a value's halves together and
`~hi ^ ~lo == hi ^ lo`, so folding the two words through it directly would give every set the same
hash as its complement — and a char class and its negation are a common pair in a compiled
matcher, so they would collide in `Optimizer.dedupOr`'s hash set. The xor-shift between the two
multiplies is load bearing as well: 2^63 is a fixed point of multiplication by any odd value mod
2^64, so without it a set holding only char 63 hashes like one holding only char 127. Over a
corpus of all 128 singletons, every contiguous range at 8 lengths, and the complement of each,
this gives 1968 distinct hashes for 1968 distinct sets.

The accompanying test walks all 128 single-character sets rather than asserting on chosen pairs.
Which pairs collide is a function of how the two words are combined, so a fixed pair silently
stops covering the defect as soon as that changes — which is exactly what happened to an earlier
revision of this PR.

`memberBit` relies on JLS 15.19 implicit shift-distance masking rather than an explicit
`c & 0x3F`: x86's `shrx` already masks the count in hardware, so the explicit form compiles to a
redundant `andl` per character. That one is easy to reintroduce, so it is called out in a comment.

The `AtlasRegistry.normalizeTags` change that was originally part of this PR is split out per
review, and will follow separately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


brharrington:
I like to keep PRs small and focused. Can we split the normalizeTags change to a separate PR?

spectator-api/src/main/java/com/netflix/spectator/impl/AsciiSet.java
```diff
@@ -238,39 +268,39 @@ private String replaceNonMembersImpl(String input, int start, char replacement)
...
+      // the same value rather than seeing a torn or stale one.
+      p = toPattern();
+      pattern = p;
+    }
+    return p;
   }
 
   @Override public int hashCode() {
-    return pattern.hashCode() + 31 * Arrays.hashCode(members);
+    return 31 * Long.hashCode(bits0) + Long.hashCode(bits1);
```
brharrington: `Long.hashCode(x) == Long.hashCode(~x)` for all `x`, so `31*Long.hashCode(bits0) + Long.hashCode(bits1)`
is complement-invariant. Measured:

- every set collides with its own complement: **258/258** (was 0/258)
- all **26/26** lower/upper letter pairs collide; `a-z` and `A-Z` both hash to `134217726`
- `none()`, `all()`, and `fromPattern("")` all hash to **0**
- distinct hashes over a 258-set corpus: **258 -> 127**

`equals` is exact so nothing is *wrong*, but this propagates through `CharClassMatcher.hashCode`
into `PatternMatcher` / `Query.Regex` and `Optimizer.dedupOr`'s hash set.

**Fix: mix before folding.**

```java
long h = bits0 * 0x9E3779B97F4A7C15L + bits1;
return (int) (h ^ (h >>> 32));
```

spectator-reg-atlas/src/main/java/com/netflix/spectator/atlas/AtlasRegistry.java
brharrington, line 475: `normalizeTags` is now too big to inline — on the hot path this PR set out to speed up

`javap -c` on both revisions: **194 -> 327 bytecodes**, and `FreqInlineSize = 325` on the JDK 25
this repo benchmarks with. It crosses by 2 bytes, so `AbstractRegistry.normalizeId` no longer
inlines it on *any* path, including the all-clean early return.

The duplicated `else` rebuild loop is ~30 of those bytecodes — deleting it gets back under, and a
split-range loop (`[1, firstDirty)` copy, `[firstDirty, size)` rebuild) is behaviorally identical
with no inner branch at all.

spectator-reg-atlas/src/jmh/java/com/netflix/spectator/atlas/NormalizeTagsBench.java
```diff
@@ -0,0 +1,79 @@
...
+      .withTag("path", "/api/v1");
+
+  /**
+   * The dirty tag is neither first nor last: earlier tags are confirmed clean and reused, but
+   * later ones were never checked by the detection scan (it breaks at the first dirty tag), so
+   * they still get rescanned by replaceNonMembers below even though they turn out to be clean.
+   */
+  private final Id dirtyMiddleTag = registry.createId("test")
+      .withTag("a", "1").withTag("b", "2").withTag("path", "/api/v1")
+      .withTag("d", "4").withTag("e", "5");
```
brharrington: `ArrayTagSet` sorts by key, so `path` sorts last in *both* benchmark ids. Printed:

```
dirtyMiddleTag actual order: test:a=1:b=2:d=4:e=5:path=/api/v1   size=6 firstDirtyTag=5
dirtyLastTag   actual order: test:a=1:b=2:c=3:d=4:path=/api/v1   size=6 firstDirtyTag=5
```

So "one dirty tag, mid-list +29%" and "one dirty tag, last +28-39%" are one case. The genuine
middle case — a clean tag sorting *after* the dirty one, i.e. the new
`i >= firstDirtyTag`-but-actually-clean branch — is **unmeasured and untested**. Use a dirty key
that sorts middle, e.g. `b=/api/v1`.

spectator-api/src/main/java/com/netflix/spectator/impl/AsciiSet.java
```diff
@@ -124,20 +160,29 @@ private static void append(StringBuilder buf, char s, char e) {
...
-    this.members = Preconditions.checkNotNull(members, "members array cannot be null");
-    this.pattern = toPattern(members);
+  /** Membership bits for characters 64-127. */
+  private final long bits1;
+
+  /**
+   * Cached, lazily computed pattern for {@link #toString()}; excluded from equals/hashCode and
+   * from serialization since it's fully derived from {@link #bits0}/{@link #bits1}.
+   */
+  private transient String pattern;
```
brharrington: `toString` isn't on the hot path, it is used for debugging and such. I think it might be cleaner to just remove this member variable and generate the pattern when needed.

spectator-api/src/main/java/com/netflix/spectator/impl/AsciiSet.java
michaelbraun: Confirmed, and it holds for the reason you gave: `Long.hashCode` is `(int)(v ^ (v >>> 32))`, i.e. `lo ^ hi`, and `~lo ^ ~hi == lo ^ hi`, so complementing both words is invisible to it. 200k random longs, zero counterexamples.

One thing to flag on the suggested fix though — mixing only `bits0` leaves half of it in place. Any set confined to chars 64-127 has `bits0 == 0`, so `bits0 * K + bits1` degenerates to `Long.hashCode(bits1)`, and that fold is symmetric between the halves. `a-z` and `A-Z` still collide at 134217726 under it.

Mixing both words with distinct odd constants fixes both kinds of pair. Over a 306-set corpus (every single char, a spread of ranges, both halves, and the complement of each):

| | distinct | set == its complement | `a-z` == `A-Z` |
|---|---|---|---|
| current | 84/306 | 153/153 | yes |
| `bits0 * K + bits1` | 250/306 | 0/153 | yes |
| `bits0 * K1 + bits1 * K2` | 300/306 | 0/153 | no |

Went with the last one. A murmur3 finalizer on top scored the same (298/306), so it didn't seem worth the extra lines.

```java
final long h = bits0 * 0x9E3779B97F4A7C15L + bits1 * 0xC2B2AE3D27D4EB4FL;
return (int) (h ^ (h >>> 32));
```

Added a test covering set-vs-complement and the half-swap pair so this can't quietly regress again.

spectator-api/src/main/java/com/netflix/spectator/impl/AsciiSet.java
michaelbraun: Agreed, removed. `toString` is only reached from `CharClassMatcher.toString` and the `replaceNonMembers` argument-validation message, so the cache was buying a mutable field and a safe-publication argument for a debug path.

Worth noting the win here wasn't from the cache: it was from getting `toPattern` out of the constructor, which the old code called eagerly on every `AsciiSet`. That's why the set-algebra operations were so slow before — `union` was paying for a `StringBuilder` walk over 128 chars per call. That part is unchanged, so the numbers hold.

spectator-api/src/main/java/com/netflix/spectator/impl/AsciiSet.java
michaelbraun: Both confirmed, and thanks — the 62/126 derivation is exactly right. `0x9E3779B97F4A7C15` ends in `0x15` = 21 ≡ 1 mod 4, so `M - 1 = 4k` and `2^62·M·(M-1) = 2^64·M·k ≡ 0`.

My validation is what let this through: I scored candidates on a distinct-hash count and never printed the colliding pairs, so "732/740 distinct" looked fine while hiding exactly this. Enumerating instead of counting immediately surfaced it.

On the suggested fix — `31 * hi32(bits0*M) + hi32(bits1*M)` — it clears 62/126 but is worse overall, 71 collisions across a 1968-set corpus against 12 for what I had. And 63/127 survives it, for a reason no constant choice fixes: **2^63 is a fixed point of multiplication by any odd value mod 2^64**, so `{63}` (bits0 = 2^63) and `{127}` (bits1 = 2^63) both produce 2^63 and summing them into one accumulator cannot separate them. The same thing happens to the 32-bit form, since `31 * 2^31 ≡ 2^31 mod 2^32`. Every candidate I tried that was "each word times an odd constant, then add" collided on 63/127, including two distinct constants and a murmur3 finalizer.

An xor-shift between the two multiplies breaks the fixed point and keeps one constant:

```java
long h = bits0 * HASH_MIXER;
h ^= h >>> 32;
h = (h + bits1) * HASH_MIXER;
return (int) (h ^ (h >>> 32));
```

Corpus of all 128 singletons, every contiguous range at 8 lengths, and the complement of each:

| | singleton collisions | corpus collisions |
|---|---|---|
| `(bits0*M + bits1)*M` | [62,126], [63,127] | 12 |
| two distinct constants | [63,127] | 14 |
| `31 * hi32` | [63,127] | 71 |
| murmur3 finalizer | [63,127] | 15 |
| xor-shift between | **none** | **0** |

1968 distinct hashes for 1968 distinct sets, so it matches the dispersion main got from hashing the whole `boolean[128]`.

brharrington (approved):

